PR #1, opened 2016-08-02
Add project settings files and basic demo app

karelhala:
The application uses both bower and npm so it can be published to both repositories.

Build requirements:
 * [node and npm](https://nodejs.org/en/download/)
 * [typings](https://github.com/typings/typings), [webpack](https://webpack.github.io/docs/installation.html)
 * [bower](https://bower.io/)

Docs requirements:
* [jsdoc](https://www.npmjs.com/package/jsdoc) installed as public package.

Build process is: 
 * `npm install` this will create vendor.js, [app-name].js and demo-app.js
 * `npm start` will build just [app-name].js and demo-app.js and run browsersync

Technology stack:
 * Typescript (tsconfig, tslint, typings), loading of application pieces is done with [import](https://www.typescriptlang.org/docs/handbook/modules.html)
   * tsconfig and tslint files are copy of same file in [hawkular-ui-components](https://github.com/hawkular/hawkular-ui-components), if there is need to some changes feel free to point them out.
 * webpack for build
  * whole build process is done with [webpack loaders](https://webpack.github.io/docs/loaders.html)
  * final application consists of `vendor.js` and `[app-name].js` using [commonschunk](https://webpack.github.io/docs/list-of-plugins.html#commonschunkplugin). The first one contains dependencies like Rx.js, lodash, angular and such. Second is application without any dependencies, this way we can load just application without dependencies (smaller application) - we should later create uglified file as well.
   * [HtmlWebpackPlugin](https://github.com/ampedandwired/html-webpack-plugin) is for creating dynamic html file for demo application.
   * [BrowserSyncPlugin](https://www.npmjs.com/package/browser-sync-webpack-plugin) for imediate changes in browser. We could use webpack's dev server, but with this we can use `npm|bower link` and see changes done in this package immediately in action.
  * angular-patternfly is sadly available only via bower [#215](https://github.com/patternfly/angular-patternfly/issues/215). So we have to use [CopyWebpackPlugin](https://www.npmjs.com/package/copy-webpack-plugin) for copying bower dependencies (this plugin will show a lot of text when building application).
 * build process does not need to use copy webpack plugin or generate `vendor.js` file, so it's moved to different webpack config, however installing and running this app requires to have these files created so this config is executed every time `npm install` is run.
 * documentation is generated via jsdoc and some angular settings: [angular-jsdoc](https://github.com/allenhwkim/angular-jsdoc) (example can be seen [here](https://rawgit.com/allenhwkim/angular-jsdoc/master/angular-template/docs/ngmap.Attr2Options.html))


karelhala:
@mtho11 could you please when this will be merged update bithound and travis badges?

mtho11:
@karelhala sure, but there shouldn't be anything blocking this PR from being merged at this point.

README.adoc
```diff
@@ -1,4 +1,4 @@
-= Hawkular Middleware Components for ManageIQ
+= Middleware Components for ManageIQ
```
himdel: This is still wrong, this repo is for *angular* components, not middleware specific. At least, neither toolbars nor gtl list are middleware in any way..

README.adoc
```diff
@@ -0,0 +1,71 @@
...
```
himdel: This is still wrong, this repo is for angular components, not middleware specific. At least, neither toolbars nor gtl list are middleware in any way..

README.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+image::https://www.bithound.io/github/hawkular/hawkular-ui-components/badges/score.svg[link="https://www.bithound.io/github/hawkular/hawkular-ui-components"]
+image::https://www.bithound.io/github/hawkular/hawkular-ui-components/badges/dependencies.svg[link="https://www.bithound.io/github/hawkular/hawkular-ui-components/master/dependencies/npm"]
+
+ifdef::env-github[]
+[link=https://travis-ci.org/hawkular/hawkular-ui-components]
+image:https://travis-ci.org/hawkular/hawkular-ui-components.svg["Build Status",
+link="https://travis-ci.org/hawkular/hawkular-ui-components"]
+endif::[]
```
himdel: Assuming these will be fixed by @mtho11  in a seaparate PR, right?

demo/index.ts
```diff
@@ -0,0 +1,22 @@
...
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+
+///<reference path="tsd.d.ts"/>
+(() => {
+  const app = angular.module('demoApp', ['miqStaticAssets', 'ui.bootstrap', 'ui.router',
```
himdel: Do we actually depend on `ui.router` here? Seems like components shouldn't...

package.json
```diff
@@ -0,0 +1,63 @@
...
+    "browser-sync-spa": "1.0.3",
+    "browser-sync-webpack-plugin": "1.0.3",
+    "copy-webpack-plugin": "3.0.1",
+    "css-loader": "^0.23.1",
+    "event-stream": "^3.3.3",
+    "extract-text-webpack-plugin": "1.0.1",
+    "file-loader": "^0.9.0",
+    "html-webpack-plugin": "^2.15.0",
+    "imports-loader": "^0.6.5",
+    "jquery": "3.0.0",
```
himdel: little thing, but jquery-3.1 maybe? Because of that whole 3.0 eats your exceptions thing..

himdel:
So.. a few issues, but nothing major really..

demo/index.ts
karelhala: The demo application is for presenting components in pure angular environment and it uses ui-router for views. I'll give it a try and see if multiple `package.json` will be good. If you want to see how the [demo folder ](https://github.com/karelhala/hawkular-ui-components/tree/master/demo) looks like.

package.json
karelhala: One thing, `jquery` package shouldn't be here at all.

demo/index.ts
himdel: Thanks! When I first commented I missed the implications of "demo",  so, try but if not it's npm not bower so, as long as that dependency doesn't propagate to the resulting bower package, no harm really :-)

karelhala:
Okay, fixed all the issues, but I think it's good to go. Last thing which is missing is probably tests, but these can be added later down the line, Mike is already working on settings for them: https://github.com/hawkular/hawkular-ui-components/pull/138

himdel:
Yeah, agreed, no sense in delaying any further.. :)

demo/index.ts
mtho11: +1 to its own package.json. That way we don't confuse the dependencies of the demo with that of the actual library.

PR #4, opened 2016-08-04
Use the current Node.js version 6.x in our dev build environment. 

mtho11:
This so we can utilize the latest ES6 in our build process without using Babel for transpiler for webpack builds in infrastructure. 

_NOTE: this only relevant to dev build and doesn't effect the runtime js components generated._

Version 5 is already deprecated and version 6.3.1 is the current version. Version 4 will be replaced with version 6 on Long Term Support(LTS) in October. The [LTS schedule](https://github.com/nodejs/LTS) is as follows:

nodejs_lts__node_js_foundation_long-term_support_working_group

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -41,7 +41,7 @@ https://www.youtube.com/watch?list=PLOETEcp3DkCq788xapkP_OU-78jhTf68j&v=AMwjDibF
 
 == Development Environment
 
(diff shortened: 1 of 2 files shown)


mtho11:
Code-wise this is what the implications mean: http://node.green/

PR #5, opened 2016-08-04
Fix Bithound badge error because of unexpected line break.

mtho11:
Unfortunately couldn't see this error until it was deployed to github.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,9 +1,7 @@
 = Angular UI Components for ManageIQ
 
-image::https://www.bithound.io/github/ManageIQ/ui-components/badges/score.svg[link="https://www.bithound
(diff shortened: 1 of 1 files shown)


himdel:
I still can't see the difference :) .. but .. looking https://github.com/mtho11/ui-components/tree/badge_linebreak_error seems like it's still not quite fixed..

I wonder if it'd be easier to just use markdown like everywhere else..

mtho11:
@himdel markdown is fine by me.

himdel:
OK, can give me a second pair of eyes on https://github.com/himdel/ui-components/tree/markdown please?

mtho11:
@himdel LGTM

PR #7, opened 2016-08-04
Install karma unit testing framework for webpack using jasmine.

mtho11:
To run tests type `npm t`

Tests are denoted as *.spec.ts in the same directory as the code (this can be easily changed if this is is not desired). This way the spec is right with the code.

specs


mtho11:
TODO: is to remove specs from production build code

PR #8, opened 2016-08-09
Toolbar explorer

karelhala:
Explorer's toolbar contains two state buttons, so we have to enable these buttons and show them properly on toolbar.
These two state buttons uses `miq-button` component and set's specific classes (activate, disabled). We have to filter out `view` items from these two state buttons since they use same toolbarItem type.

New separators inside dropdown buttons.

Enable strict mode in webpack so the `npm` commands will not throw errors [Strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode).

### DIFF
--- a/dist/js/ui-components.js
+++ b/dist/js/ui-components.js
@@ -44,18 +44,13 @@
 /* 0 */
 /***/ function(module, exports, __webpack_require__) {
 
-	__webpack_require__(15);
-	module.exports = __webpack_require__(17);
+	__webpack_require__(14);
+	module.exports = __webpack_require__(16);
 
 
 /***/ },
 /* 1 */,
-/* 2 */
-/***/ function(module, exports) {
-
-	module.exports = angular;
-
-/***/ },
+/* 2 */,
 /* 3 */,
 /* 4 */,
 /* 5 */,
@@ -67,22 +62,21 @@
 /* 11 */,
 /* 12 */,
 /* 13 */,
-/* 14 */,
-/* 15 */
+/* 14 */
 /***/ function(module, exports) {
 
 	// removed by extract-text-webpack-plugin
 
 /***/ },
-/* 16 */,
-/* 17 */
+/* 15 */,
+/* 16 */
 /***/ function(module, exports, __webpack_require__) {
 
-	/* WEBPACK VAR INJECTION */(function(angular) {"use strict";
+	"use strict";
 	///<reference path="tsd.d.ts"/>
-	var services_1 = __webpack_require__(18);
-	var components_1 = __webpack_require__(21);
-	var filters_1 = __webpack_require__(32);
+	var services_1 = __webpack_require__(17);
+	var components_1 = __webpack_require__(20);
+	var filters_1 = __webpack_require__(30);
 	var miqStaticAssets;
 	(function (miqStaticAssets) {
 	    miqStaticAssets.app = angular.module('miqStaticAssets', ['ui.bootstrap', 'ui.bootstrap.tabs', 'rx', 'ngSanitize']);
@@ -91,15 +85,14 @@
 	    filters_1.default(miqStaticAssets.app);
 	})(miqStaticAssets || (miqStaticAssets = {}));
 
-	/* WEBPACK VAR INJECTION */}.call(exports, __webpack_require__(2)))
 
 /***/ },
-/* 18 */
+/* 17 */
 /***/ function(module, exports, __webpack_require__) {
 
 	"use strict";
-	var endpointsService_1 = __webpack_require__(19);
-	var toolbarSettingsService_1 = __webpack_require__(20);
+	var endpointsService_1 = __webpack_require__(18);
+	var toolbarSettingsService_1 = __webpack_require__(19);
 	Object.defineProperty(exports, "__esModule", { value: true });
 	exports.default = function (module) {
 	    module.service('MiQEndpointsService', endpointsService_1.default);
@@ -108,7 +101,7 @@
 
 
 /***/ },
-/* 19 */
+/* 18 */
 /***/ function(module, exports) {
 
 	"use strict";
@@ -136,7 +129,7 @@
 
 
 /***/ },
-/* 20 */
+/* 19 */
 /***/ function(module, exports) {
 
 	"use strict";
@@ -245,11 +238,11 @@
 
 
 /***/ },
-/* 21 */
+/* 20 */
 /***/ function(module, exports, __webpack_require__) {
 
 	"use strict";
-	var toolbar_menu_1 = __webpack_require__(22);
+	var toolbar_menu_1 = __webpack_require__(21);
 	Object.defineProperty(exports, "__esModule", { value: true });
 	exports.default = function (module) {
 	    toolbar_menu_1.default(module);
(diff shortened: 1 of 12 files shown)


karelhala:
No tests and no travis config, yet. This PR is for adding toolbar's explorer support, but thanks to @mtho11 we are now able to add tests so I'll do this (travis config + toolbar tests) in next PR.

src/components/toolbar-menu/toolbar-menu.html
```diff
@@ -19,3 +23,6 @@
                       class="miq-view-list">
     </miq-toolbar-view>
 </div>
+
+
+<!--<button data-url_parms="?compare_task=same&amp;id=" data-url="compare_miq_same" title="Attributes with same values" id="compare_same" data-click="compare_same" name="compare_same" type="button" class="btn btn-default"><i class="product product-compare_same fa-lg" style=""></i>&nbsp;</button>-->
```
himdel: ;)

src/components/toolbar-menu/toolbar-menu.html
```diff
@@ -6,6 +6,10 @@
                           toolbar-button="item"
                           on-item-click="vm.onItemClick(item, $event)">
       </miq-toolbar-button>
+      <miq-toolbar-button ng-repeat="item in toolbarItem | filter: {type: 'buttonTwoState'}:true | filter: {id: '!view'}"
```
mtho11: How about some custom filters here? Maybe something like 'filterByStateAndView'.

src/components/toolbar-menu/toolbarComponent.ts
```diff
@@ -106,7 +106,12 @@ export class ToolbarController {
    * @returns {boolean} true|false if it's item with button or button select type.
    */
   private static isButtonOrSelect(item: IToolbarItem): boolean {
-    return item.type && ToolbarController.isButtonSelect(item) || ToolbarController.isButton(item);
+    return item.type && (ToolbarController.isButtonSelect(item) || ToolbarController.isButton(item) ||
+      ToolbarController.isButtonTwoState(item));
+  }
+
+  private static isButtonTwoState(item: IToolbarItem): boolean {
+    return item.type === 'buttonTwoState';
```
mtho11: Should item.type be an `enum`?

src/components/toolbar-menu/toolbar-menu.html
karelhala: Good catch :+1:

karelhala:
@mtho11 thanks for poiting out the filter and enum, looks much cleaner now.

@himdel please recheck this, thank you.

src/components/toolbar-menu/toolbarComponent.ts
```diff
@@ -73,6 +75,36 @@ export class ToolbarController {
   }
 
   /**
+   * Helper method for getting string value of {@link ToolbarType.BUTTON_SELECT}
+   * @memberof ToolbarController
+   * @function getToolbarListType
+   * @returns {string}
+   */
+  public getToolbarListType(): string {
+    return ToolbarType.BUTTON_SELECT.toString();
```
himdel: honestly, I don't know... enums sound like a great idea but if this is what it takes, I'm starting to doubt the value

himdel:
@martinpovolny what do you thing of the latest changes here please? I wonder if this is becoming too java-like..

src/components/toolbar-menu/toolbarComponent.ts
karelhala: This is just to access toolbarType enum from teplate, but I get it this is more [java based enum](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) we can always use [typescript's way](https://www.typescriptlang.org/docs/handbook/enums.html), cleaner and less code.

mtho11: I was thinking the just a typescript enum: https://www.typescriptlang.org/docs/handbook/enums.html

mtho11:
@himdel @martinpovolny I think the typescript enum addresses your concerns from above?

himdel:
I think so yes, seems like it doesn't require manually creating methods for each case and calling toString everywhere, so yes :)

martinpovolny:
@himdel : I don't mind "java-like" but I'd love to aim for some predictability and consistency in the APIs

karelhala:
I've changed it to more cleaner way, but as suggested [here](http://stackoverflow.com/questions/15490560/create-an-enum-with-string-values-in-typescript) the problem is if we want to do something like: `var test:E = E.hello`, then we'd have to use enum with class, constructor and toString method. Here it's not required, but so we know this in future.

himdel:
@karelhala I think the stringly enum mentioned in the same link is exactly what we should use instead..

Right now, you're adding 70 lines of code to do the job of 
```
type ToolbarType = "button" | "buttonTwoState" | "buttonSelect" | "custom";
```

:)

karelhala:
That's type definition, it would be nice if there will be option to target different types, so calling something like `ToolbarType.button` will not work.

We can use this, but we'd have to nonetheless use some class to store these values
```Typescript
type TToolbarType = "button" | "buttonTwoState" | "buttonSelect" | "custom";
class ToolbarType {
  public static BUTTON = 'button';
  public static BUTTON_TWO_STATE = 'buttonTwoState';
  public static BUTTON_SELECT = 'buttonSelect';
  public static CUSTOM = 'custom';
}
```

And then  calling `ToolbarType.BUTTON`, but that's the same as I wrote (ok I wrote too much JSOC comments, will remove them)

himdel:
Aah, thanks! Yeah, that should works too, suddenly it's much more readable :)

dist/js/ui-components.js
```diff
@@ -623,6 +657,78 @@
...
+	 * ToolbarType.BUTTON_TWO_STATE;
+	 * ToolbarType.BUTTON_SELECT;
+	 * ToolbarType.CUSTOM;
+	 * ```
+	 * It will return string interpretation of toolbarType.
+	 * @memberof miqStaticAssets
+	 * @ngdoc enum
+	 * @name ToolbarType
+	 */
+	var ToolbarType = (function () {
```
himdel: @karelhala We may have a problem here, have you tested this please?

The ts code looks good and I'd merge .. but seems like in the compiled JS, `ToolbarType.BUTTON_TWO_STATE` will always be `undefined` (and the others..)

karelhala: Yeah, looks like I got outsmarted by typescript :smile:

himdel:
@martinpovolny you're testing these changes by testing #9753 right? I'm OK with merging this if it works for you..

martinpovolny:
right:
```
devil - [~/Projects/manageiq-review] (pull-request-karelhala-toolbarAngular)$ git grep ui-compo
...
> bower.json:    "ui-components": "karelhala/ui-components#toolbar_explorer"
```

martinpovolny:
@karelhala : feel free to merge it if it's convent for you. Or do you want to add more fixes to this PR?

mtho11:
@martinpovolny we should probably merge now before its gets any larger. We can always add smaller fixes later.

PR #10, opened 2016-08-16
Add image to toolbar list

karelhala:
When toolbar item does not have icon, we use img_url of this item.

Add $event to view change click.


karelhala:
We should update bower, npm version and create tag so we can reference these changes as `0.0.2` instead of branch name.

dist/js/ui-components.js
```diff
@@ -301,6 +341,7 @@
 	        this.$window = $window;
 	        this.$location = $location;
 	        this.$sce = $sce;
+	        console.log(this.toolbarItems);
```
himdel: Mind regenerating the js? Pretty sure you must have added this manually..

karelhala: Yeah removed the console.log but, forgot to rebuild the project before pushing. Good catch.

himdel:
LGTM ,merging and releasing 0.0.2..

PR #11, opened 2016-08-17
Reset countSelected after new toolbar object

karelhala:
When new toolbar is created we have to reset counter for number of selected items.


himdel:
Fixes [the problem](https://github.com/ManageIQ/manageiq/pull/9753#issuecomment-240404780), merging

PR #13, opened 2016-08-24
Make separators and toolbar list items hidable

karelhala:
@jzigmund proposed that toolbar separators and toolbar items could be hidable based on hidden property.
Toolbar [json](https://paste.fedoraproject.org/413324/46446147/) to reproduce
Old way:
selection_059

EDIT: if toolbar list does not contain any visible items, make whole list invisible.

PR #14, opened 2016-08-25
Add image icon to toolbar simple buttons

karelhala:
When button has no icon, but image_url is present use this image_url instead of icon as img tag

### DIFF
--- a/dist/js/ui-components.js
+++ b/dist/js/ui-components.js
@@ -596,7 +596,7 @@
 /* 26 */
 /***/ function(module, exports) {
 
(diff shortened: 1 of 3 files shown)


karelhala:
@mtho11 after mergin this and https://github.com/ManageIQ/ui-components/pull/13 please release new version of `ui-components`

karelhala:
As proposed by @skateman and @himdel MiQ will move soon-ish to `patternfly-sass: 3.9.0`, this will break toolbars so they will show switching of views right next to toolbar elements. So I've updated toolbar-view component so we don't have to change anything after migrating to `3.9.0`

himdel:
Verified that with new patternfly, the view toolbar buttons are correctly pulled right while `download_choice` remains on the left side.. :+1:

himdel:
Available in 0.0.6

PR #15, opened 2016-08-31
Enable text in simple toolbar buttons

karelhala:
There was no text in simple toolbar buttond, but with introduction of https://github.com/ManageIQ/manageiq/pull/10648 we need to show text in these buttons as well. So show them after image / icon.


mtho11:
@karelhala as you fix issues please add tests to prove the issue before/after. It will save our sanity later (especially for other maintainers).

PR #16, opened 2016-09-01
Toolbar tests

karelhala:
Full test cases for toolbar components.

If you need to split this into couple of PRs (one for toolbar button, other for toolbar view, etc.) let me know.

25 tests are passing and are in line with newest requirements in MiQ.

After merging this issue we can create `travis.yml` file with travis settings.


himdel:
`src/components/toolbar-menu/toolbarButton.json` .. the name looks a bit misleading, assuming it's only ever used for testing, it should be somehow obvious.. `src -> spec`? Or `toolbarButton.sample.json`?

karelhala:
>  .. the name looks a bit misleading, assuming it's only ..

That's true, I'll rename it to `toolbarButton.sample.json` since we keep tests in same folder as sources.

himdel:
Thanks, and I guess we're not using `ui.bootstrap` at all here? And sanitize was being used but not depended on?

If so, I'll just merge :)

karelhala:
Yes exactly, I think keeping this angular module as slim as possible (no strict dependencies while bootstrapping this) is good idea.

Best example can be done with not including `ng-patternfly`, if some angular application will include this module, but will not include `ng-patternfly` such application will not be able to see patternfly's components (for example [tile view](http://angular-patternfly.rhcloud.com/#/api/patternfly.views.directive:pfCardView)), but still can use toolbars (they do not depend on patternfly at all)

himdel:
Sure, makes sense to only include stuff we really need for the core functionality :).

Merging..

himdel:
I guess we don't need a new release just for this change, right?

mtho11:
No, I don't think we need a release for this testing stuff.

PR #17, opened 2016-09-06
ToolbarView - update html to match patternfly 3.10

himdel:
Without this, all the view toolbar items are blue, all the time, as of ManageIQ/manageiq#11020

Updated the view toolbar to match https://www.patternfly.org/pattern-library/content-views/list-view/

Cc: @karelhala, @skateman, @martinpovolny 

You can test by doing `bower install manageiq-ui-components=himdel/ui-components#fix-pf-310` under `manageiq/`

### DIFF
--- a/dist/js/ui-components.js
+++ b/dist/js/ui-components.js
@@ -44,8 +44,8 @@
 /* 0 */
 /***/ function(module, exports, __webpack_require__) {
 
-	__webpack_require__(18);
-	module.exports = __webpack_require__(20);
+	__webpack_require__(14);
+	module.exports = __webpack_require__(16);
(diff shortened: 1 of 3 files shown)


skateman:
Verified fix! :+1:

karelhala:
@himdel looks good! Much better than using `<ul>` and `<li>` could you try running `npm t` to see if test are fine? Looks like `toolbarViewComponent.spec.ts:42` should be changed to expect buttons.

himdel:
Updated the specs, now they seem to pass :)

himdel:
Released 0.0.7..

PR #18, opened 2016-09-08
[WIP] Gtl components

karelhala:
This is PR for enabling GTL components in MiQ.

General question, should I split this into two seperated PRs? Since this is adding table and tile views. However these two parts cooparete together.

For some strange reason I found that demo page is slow with filtering tile components. Will investigate why it's so slow, it's not happening in MiQ so it should not happen here as well.

I will add pictures with what it looks like later on.

TODO:
- [ ] add pictures
- [ ] cleanup
- [ ] add tests
- [ ] add comments


karelhala:
Definetley splitting this PR, even without dist files and JSON file for datatbles it's still 600 LOC.

himdel:
Since you'll be splitting this one, what do you think about doing the directory structure a bit differently?

Right now, we have `components/`, `services/`, `filters/`, `interfaces/`, `styles/`, where everywhere except for components, unrelated things are just thrown together in a big pile..

I'd much rather see something like `src/toolbars` (which gives you, say, a `miq.toolbars` module), `src/gtl` (which gives you `miq.gtl` containing all the three GTL components and their related services and styles), ...etc (and maybe some `src/common` for stuff that are shared between these modules, if there's any).

karelhala:
Great idea, I will keep this PR open until whole module is merged.

PR #19, opened 2016-09-12
Split application to angular modules

karelhala:
As proposed by @himdel in https://github.com/ManageIQ/ui-components/pull/18 it would be convenient to split this application to more specific angular modules, so end application can include only one module. Also it is much easier to maintain smaller modules, rather than one big.

Currently there are two modules:
 * common - services and components which does not belong to any specific module
 * toolbar - this module contains services and components for toolbar section only

There is also present styles file which contains all style settings for whole application.

Both these new modules are combined into one module.

If we will want to add new module in future we'll add name of this module to `application-settings.js` with given path (inside `src` folder) to this module. Don't forget to re-run `npm start` since these settings are not automatically presented in webpack processes.

`npm t` shows no errors, also running this in ManageIq main application works fine.


karelhala:
The one big module is open for discussion.
Benefits - we do not need to load each module in final application.
Drawbacks - this angular's module will be empty and will just load small modules.

himdel:
Ad one big module.. I think it's quite OK to have one even if it just brings all the modules together, I assume ManageIQ will always depend on *all of them* anyway.. Except for parts of miq which may ask for just specific modules.. :)

karelhala:
> One  module to bring them all and in the darkness bind them

sorry I couldn't resist :smile: 

I've changed tests to mock only specific module (`angular.mock.module('miqStaticAssets.toolbar')`) so if it looks good and @mtho11 has no objections either we can merge it. Not sure if new version is needed, might be better to release after GTL is done as well.

mtho11:
No objections from me. Merge.

PR #20, opened 2016-09-12
Add travis build file with build script

karelhala:
New travis build config file, we now just have to enable travis builds for ui-components.

This travis build is trivial, specifies node and npm versions, before running build script it will install necessary dependencies.

Build script will just install this application and runs tests for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+sudo: required
+dist: trusty
+language: node_js
+node_js:
+  - '5'
+
+npm:
+  - '3'
+
+cache:
+  directories:
+    - node_modules
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - '3'
+
+cache:
+  directories:
+    - node_modules
+
+before_install:
+  - 'npm install -g typings bower webpack karma typescript'
+
+script: ./build.sh
```
himdel: `./build.sh` != `.build.sh` ;)

himdel:
Looks reasonable, merging :)

PR #24, opened 2016-09-16
Gtl module

karelhala:
Start of gtl module, this module will contain table, grid and list components.

This initial PR introduces services for these components (fetching data from URL) and adds this module to main module.


src/gtl/services/dataTableService.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Static function which will generate http get config from given variables.
+   * @param modelName string with name of model.
+   * @param activeTree string with active tree.
+   * @param currId ID of current item.
+   * @returns {{params: {}}} config object with params set.
+   */
+  private static generateConfig(modelName?: string, activeTree?: string, currId?: string) {
+    let config = {params: {}};
+    _.assign(config.params, DataTableService.generateModelConfig(modelName));
```
mtho11: We probably need to figure out the priority of libraries to use, when there is overlapping functionality. This statement is totally correct but should one use Object.assign(), _.assign(), [or  even angular.merge, angular.extend].

Anyway, we probably need a standard approach to this so that each of of doesn't do it differently.
@himdel @martinpovolny wdyt?

mtho11:
@karelhala btw I'm glad the loader.ts has changed to index.ts seems much more standard.

karelhala:
New tests were added, so I think it's good to go. WDYT @himdel @martinpovolny @mtho11

src/gtl/services/dataTableService.ts
himdel: I'd go in the order of [ES6, lodash, angular, jquery], except for when the angular functions are explicitly needed because we want to skip stuff like `$id` etc.

himdel: ...but then again, that (es6 over lodash) assumes es6-shim or new browsers. We do that in miq, but not sure how far to take the library thing..

karelhala: I'd be more benevolent with this and put ES6 and lodash at the same level. But then egain, we are using Typescript and we no longer need to use `_.map`, `_.each` but use ES6's equivalents.

But if we agree on using more ES6 things over lodash's I'm ok with it.

himdel: > I'd be more benevolent with this and put ES6 and lodash at the same level.

No problem with that :). Either way, I guess there's no reason to change this PR..

PR #29, opened 2016-09-21
Enable coveralls

karelhala:
Tests are running, however there was no coverage report tool, so enable [coveralls.io](https://coveralls.io/github/ManageIQ/ui-components)

[karma coverage](https://karma-runner.github.io/0.8/config/coverage.html) is used for generating coverage data.

Current code coverage:
selection_073

PR #30, opened 2016-09-21
coverage.sh - cd to scripts before assuming ../node_modules

himdel:
Introduced in #29, `coverage.sh` is not run with CWD=scripts/, causing it to fail - fixed by changing to the right directory first..

@karelhala can you..? :)


karelhala:
Thanks for fixing it.

PR #31, opened 2016-09-22
Gtl table

karelhala:
New component with name miqDataTable `miq-data-table` was added.
* [x] update docs
* [x] tests for controller
* [x] tests for component

selection_074

PR #32, opened 2016-09-23
Gtl list

karelhala:
This PR will add 2 new component smiqTileVie `miq-tile-view` it's similar to datatable, so it extends same abstract class.
And miqPaging `miq-paging` used for pagination in tile view.
- [x] tests for pagination (both controller and component)
- [x] tests for tile (both controller and component)
- [x] refactor jsdoc for tiles and pagination
### Small tile (type set to `small`):

selection_075
### Big tile (type set to `big`):

selection_076

No qaudicons, because logic behind quadicons is hard to move to JS which will be placed in there.


karelhala:
Not sure why there's so much commits, but it might be because this branch is rebased on top of gtl_table, so first merge https://github.com/ManageIQ/ui-components/pull/31

karelhala:
No longer WIP, altho there is bug in this repository when changing page number rewrite of items is really slow. Don't know why, but it does not happen in MiQ main repository.

karelhala:
Caught the bug with not refreshing immediately on changing pages. Fault is either on Patternfly or Angular (would bet more to Angular), because transclude with repeated items is not satisfactory and there is couple of seconds delay.

Hot fix for this is to rewrite patternfly's implementation of pf-card view without transclusion.

demo/controllers/dataTableController.ts
```diff
@@ -30,6 +31,8 @@ export default class DataTableController {
       let start = (this.settings.current - 1) * this.settings.perpage;
       this.onLoadNext(start, this.settings.perpage);
       this.onSort(2, true);
+      this.settings.selectAllTitle = 'Select All';
```
himdel: We'll need to i18n this, is there a way to override these from the outside?

src/gtl/components/tile-view/tileViewComponent.ts
```diff
@@ -0,0 +1,186 @@
...
+   * @memberof TileViewController
+   * @function onTileClick
+   * @param item which tile was clicked.
+   */
+  public onTileClick(item) {
+    this.onItemSelected({item: item, isSelected: item === _.find(this.options.selectedItems, {id: item.id})});
+  }
+
+  public onTileSelect(item) {
+    console.log(item);
```
himdel: stray console.log :)

himdel:
Also will need rebasing I guess .. and the first 2 commits look like leftovers from a previous PR maybe?

demo/controllers/dataTableController.ts
karelhala: These are in demo app, you have to send `selectAllTitle` and `sortedByTitle` from outside app. This way we can use i18 in MiQ.

But we should definitely add i18 to this.

karelhala:
@himdel, @mtho11, @martinpovolny all done, you can review changes and merge this branch.

martinpovolny:
Ping @dclarizio, @epwinchell : I'd like to bring this to your attention. In the review of this PR I'd like to focus on basic functionality and solve UX as well as any details in follow-up PRs.

I'd like to get this in early in the dev cycle for the next release.

epwinchell:
@martinpovolny @karelhala Is the idea of this PR to replace the Grid View or Tile View with PF Cards?

As discussed in https://github.com/ManageIQ/manageiq/pull/11114 with @serenamarie125 , we still need to sort out what the new GTL will actually include (possible replacement of tile view with PF List View, fixed paging bar vs in-view (as you've done above), paging vs infinite scrolling, etc)

karelhala:
The replacement of Grid and Tile with PF Cards is just convenient, since I replaced whole report data (GTL), I though that adding PF cards will be good so the users will see change and they are more responsive and feel more UX friendly than what is in MiQ.

However I wouldn't hold this back for much longer. The optimal way (at least in my opinion) is to merge this quite soon (along side with https://github.com/ManageIQ/manageiq/pull/11114) and work on changes in this repo. UI/UX changes will be much easier here than in MiQ.

martinpovolny:
I agree with @karelhala. We need to move forward cannot do everything at once.

PR #33, opened 2016-10-11
toolbarSettingsService - add setCount method, to replace checkboxClicked

himdel:
The toolbar needs to know how many items are currently selected to be able to enable/disable items with the `onwhen` attribute.

Previously, this was done by calling checkboxClicked which increments/decrements the counter, but that can sometimes lead to inconsistent state (like toolbarCount going to -2).

Adding a method to simply set the counter, since we already do the counting in miq anyway.

Cc @karelhala , @martinpovolny


himdel:
Released v0.0.8

PR #34, opened 2016-11-02
Fix a typo in IToolbarData.

martinpovolny:
just a small typo


himdel:
No other `targer` references, it's an interface fix so no need to regenerate js or do a new release.. merged :)

PR #36, opened 2016-11-16
Npm package

himdel:
Cleaned up dist/, changed the name in package.json to manageiq-ui-components and the version to match bower.

Also updated the wiki instructions on building a new version.

So, we should now have a manageiq-ui-components 0.0.9 npm package, I'll add you @martinpovolny and @karelhala to the owner list so that you can push too, please give me the login you use for npm..


karelhala:
Looks good to me.

Also add @mtho11 to this npm package.

martinpovolny:
my account: martin_povolny

himdel:
added :)

PR #37, opened 2016-11-21
Add right margin to toolbar button icons

petervo:
Matches margin used in toolbarList buttons

### DIFF
--- a/src/toolbar/components/toolbar-menu/toolbar-button.html
+++ b/src/toolbar/components/toolbar-menu/toolbar-button.html
@@ -12,7 +12,8 @@
         ng-class="{active: toolbarButton.selected, disabled: !toolbarButton.enabled}"
         ng-hide="toolbarButton.hidden"
         ng-click="onItemClick({item: toolbarButton, $event: $event})">
-  <i ng-if="toolbarButton.icon" class="{{toolbarButton.icon}}" style=""></i>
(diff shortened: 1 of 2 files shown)


martinpovolny:
ping @epwinchell

epwinchell:
@petervo Let's avoid inline styling. For the time being, toolbar styling is still handled here: https://github.com/ManageIQ/manageiq/blob/master/app/assets/stylesheets/patternfly_overrides.scss#L388

petervo:
Just copying the current implementation.

https://github.com/ManageIQ/ui-components/blob/master/src/toolbar/components/toolbar-menu/toolbar-list.html#L4

But feel free to close and fix however you prefer. It just doesn't look great right now.

epwinchell:
@petervo Please include a screenshot. Thanks

himdel:
So, @petervo this only affects first-level buttons, right?

What I'm seeing is that buttons with text do look better...

before

ui37z

after

ui37y


But buttons without any text look worse...

before

ui37x

after

ui37


Care to fix? :)

petervo:
@himdel, yep good point. Fixed, but as @epwinchell said it's probably better / more correct to use css selectors or classes for this but i don't really have time to dig into that right now. So feel free to close and fix another way if this is too hacky.

himdel:
@petervo I'd definitely have preferred css classes but.. well, this is still better and you're right that it's not the only inline style we need to fix there.

So .. merging .. and creating an refactoring issue :)

Thanks

PR #38, opened 2016-11-25
npm - use prepublish instead of postinstall

himdel:
postinstall is not only called after `npm install`, but also after having installed the package

prepublish is called only before publish *and* on doing a standalone `npm install`

So moving those actions to prepublish so that people using the package won't need typescript & typings, etc. Also removed bower install, because we can't assume that people using the npm package will even have bower..

I've already released a npm version of the package to be able to test, I'll do the bower thing after this is merged...

Cc @karelhala - any other thoughts where to put it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "manageiq-ui-components",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "description": "UI components for ManageIQ project",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


karelhala:
Good one, could you please add one task which will install bower dependencies? So when someone is developing this he gets Patternfly?

himdel:
@karelhala uh.. sure :) So that you can run `npm run bower` instead of having to type out `bower install`? :)

karelhala:
Yes, and no need to have bower installed globally, but from `node_modules/.bin/bower`.

himdel:
@karelhala Allright, added :)

PR #42, opened 2016-12-02
Add site switcher for use in ops, admin and server ui's

dgutride:
This adds a dropdown for the banner that allows for switching between new ui's.  A new home landing page will be added as well allowing for a centralized landing page.

New unit tests related to the switcher have also been added and ts-formatter has been run for these changed files.

screenshot-localhost 4000 2016-12-02 11-04-59


dgutride:
@karelhala - Who should I tag that can review this?

karelhala:
I can look at this, next time you can tag me, @himdel or @mtho11.

src/site-switcher/site-switcher.html
```diff
@@ -0,0 +1,13 @@
+<div class="dropdown miq-siteswitcher">
+  <button class="btn btn-link dropdown-toggle" type="button" id="domain-switcher" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+    <span class="fa fa-lg fa-th miq-siteswitcher-icon"></span>
+  </button>
+  <div class="dropdown-menu dropdown-menu-right" aria-labelledby="domain-switcher">
+    <div ng-repeat="site in ctrl.sites" class="miq-siteswitcher-entry">
+      <a class="miq-siteswitcher-link" href="{{site.url}}" target="_blank">
+        <span title="{{site.tooltip}}" class="fa fa-3x {{site.iconClass}}"></span>
```
karelhala: Please use `ng-class` instead of `class="{{iconClass}}", I know that it's not this case now, but we might send in future and object or array and it would be broken.
```html
...
<span title="{{site.tooltip}}" class="fa fa-3x" ng-class="site.iconClass"></span>
...
```

src/site-switcher/site-switcher.html
```diff
@@ -0,0 +1,13 @@
+<div class="dropdown miq-siteswitcher">
+  <button class="btn btn-link dropdown-toggle" type="button" id="domain-switcher" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+    <span class="fa fa-lg fa-th miq-siteswitcher-icon"></span>
+  </button>
+  <div class="dropdown-menu dropdown-menu-right" aria-labelledby="domain-switcher">
+    <div ng-repeat="site in ctrl.sites" class="miq-siteswitcher-entry">
+      <a class="miq-siteswitcher-link" href="{{site.url}}" target="_blank">
```
karelhala: Use `ng-href` instead of `href`, there might be a problem that URLs will be fetched asynchronously and user will click on such URL and it will lead him to 404.
```html
<a class="miq-siteswitcher-link" ng-href="{{site.url}}" target="_blank">
```

src/styles/ui-components.scss
```diff
@@ -210,3 +210,95 @@ img.miq-gradient-background {
...
+.miq-siteswitcher-entry {
+  display: inline-block;
+  padding: 10px;
+  text-align: center;
+  min-width: 90px;
+}
+
+* {
+  //outline: 1px solid red;
+}
```
karelhala: Empty rule.

src/styles/ui-components.scss
```diff
@@ -210,3 +210,95 @@ img.miq-gradient-background {
...
+
+.miq-siteswitcher-entry {
+  border-color: #fff;
+  border-style: solid;
+  border-width: 1px;
+  a {
+    //color: black;
+  }
```
karelhala: Empty rule.

src/site-switcher/siteSwitcherComponent.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Controller for site switcher component
+ * @memberof miqStaticAssets
+ * @ngdoc controller
+ * @name SiteSwitcherController
+ */
+export class SiteSwitcherController {
```
karelhala: Please add `sites` attribute to this class, and might be good to add some interface to describe what properties is needed in site, no need to create custom file, just add it above controller's class.
```typescript
export interface Site {
  url: string;
  site: string;
  tooltip?: string;
  iconClass? any;
}
```

dgutride:
@karelhala  - thanks for the review, all requested changes have been made

karelhala:
Looks good, if @mtho11 or @himdel has no objections we can merge this.

src/site-switcher/siteSwitcherComponent.ts
mtho11: Personally, I'm not wild about the `any` in a public interface. I don't think private methods need any typing, but public interfaces are where we are trying to control the interchange of data amongst various clients (internal and external). For instance, we shouldn't allow `1` to be passed to iconClass. Ideally, it would be an `enum` of our iconClasses that would instantly pop-up as we type the iconClass so that invalid icons would be caught at compile time not runtime. 

The enums don't need to be in this PR, but just thoughts for later. Unless people don't agree with these enums typing at all? Which we should discuss now.

mtho11: Another addition example: 
https://github.com/hawkular/hawkular-ui/blob/8ce9c4b744477553c306a361c691dbc0b7f0b687/src/main/scripts/plugins/metrics/ts/types/metricsTypes.ts

PR #41, opened 2016-11-29
Remove typings dependencies

karelhala:
Since we are using typescript 2 we don't need to use typings (we should stop using them and move to supported @types at npm [future of declarative files](https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/))

When using new module we have to add it to `webpack.config.js` to externals object:
```javascript
{"nameWeUseInApp": "nameOfLirary"}
```
Great example is lodash, `window` object has lodash stored as `_`, but when loading this module we want to call `lodash` (`import * as _ from 'lodash'` looks better than `import * as _ from '_'`).

PR #43, opened 2016-12-08
Pagination top

karelhala:
Pagination has been groupped together and moved only to top part of page. This is useful for main MiQ repo, where we want it to be at bottom, moving it to bottom is much easier (only some css styling).

In `miq-toolbar-list` component which is for showing dropdown items added new field `dropDownClass`, we can send additional classes for dropdown element (can be array or string).

In `miq-sort-items` component which is wrapper component for `pfSort` added field `dropDownClass`, similiar to `miq-toolbar-list` but here we apply classes directly to DOM element (accessed from controller's `$element.find('.dropdown')`), however we have to wait for DOM to be rendered so I've added timeout with no time set. Same as with `miq-toolbar-list` we can send either array or string.

Field `dropDownClass` has no effect in this repo, but it's mandatory so we can change position of dropdown by sending different class in settings parameter for each table and tile.
```javascript
settings.dropDownClass = ['dropup'];
```

selection_101
selection_102


karelhala:
Just FYI this should be merged before https://github.com/ManageIQ/ui-components/pull/41 since it still uses typings.

martinpovolny:
Not sure why the tests are failing. Restarted the job in travis...

karelhala:
@martinpovolny my bad, I forgot to remove tests which were related to table header and footer. It's fixed now.

PR #46, opened 2016-12-15
Fix pagination on mobile, change height of cards ro 175px instead 100%

karelhala:
Remove height of cards to have `175px`, consistent height of cards instead of different heigh (if content is un-even in different cards).

When switching to mobile version (or when browser window was below 900px) pagination was broken, so this will fix it and show it same as in full window. Same applies to cards.


src/gtl/components/tile-view/tile-view.html
```diff
@@ -1,22 +1,22 @@
 <div class="miq-tile-section">
   <div ng-if="tileCtrl.settings.isLoading" class="spinner spinner-lg"></div>
   <div class="row miq-pagination" ng-if="tileCtrl.rows && tileCtrl.rows.length !== 0">
-    <div ng-if="tileCtrl.rows.length !== 0" class="miq-select-all col-md-3 col-ld-3">
+    <div ng-if="tileCtrl.rows.length !== 0" class="miq-select-all col-md-3 col-ld-3 col-xs-3">
       <label>{{tileCtrl.settings.selectAllTitle}}: </label>
       <input type="checkbox" ng-model="isChecked" ng-click="tileCtrl.onCheckAllTiles(isChecked)" title="{{tileCtrl.settings.selectAllTitle}}" />
     </div>
-    <div class="miq-per-page col-md-2 col-ld-2">
+    <div class="miq-per-page col-md-2 col-ld-2 col-xs-3">
```
himdel: `xs-3` but `md-2` .. indented or a typo? (Same on line 14)

Also, `col-ld-2` doesn't do anything, it's supposed to be `lg` I think ;) (all over this file..)

karelhala: `xs-2` was too small, and used 2 rows when there was item with label, I'll give it a try and use `col-md-3` and `col-lg-3`.

Yeah my bad about `ld`, I thought bootstrap is consistent at least in medium and large, looks like I was wrong...

karelhala:
ping @himdel @mtho11

src/gtl/components/tile-view/tile-view.html
```diff
@@ -1,25 +1,25 @@
 <div class="miq-tile-section">
-  <div ng-if="tileCtrl.settings.isLoading" class="spinner spinner-lg"></div>
-  <div class="row miq-pagination" ng-if="tileCtrl.rows && tileCtrl.rows.length !== 0">
-    <div ng-if="tileCtrl.rows.length !== 0" class="miq-select-all col-md-3 col-ld-3">
-      <label>{{tileCtrl.settings.selectAllTitle}}: </label>
-      <input type="checkbox" ng-model="isChecked" ng-click="tileCtrl.onCheckAllTiles(isChecked)" title="{{tileCtrl.settings.selectAllTitle}}" />
+    <div ng-if="tileCtrl.settings.isLoading" class="spinner spinner-lg"></div>
```
mtho11: Should ARIA attributes be added for accessibility?
Probably a separate PR if true.

^^ @martinpovolny @himdel

PR #48, opened 2016-12-20
Yarn replacement of npm

mtho11:
Yarn provides a faster more reliable (deterministic via yarn.lock) build replacement for npm while still using npm packages.

#### Resources:

- [Migrating from npm to yarn](https://yarnpkg.com/en/docs/migrating-from-npm)

- [Yarn CLI](https://yarnpkg.com/en/docs/cli/)

- [NPM vs Yarn Cheatsheet](https://shift.infinite.red/npm-vs-yarn-cheat-sheet-8755b092e5cc#.8pmpnu9cq)

- [Yarn Package Manager an Improvement over NPM](https://scotch.io/tutorials/yarn-package-manager-an-improvement-over-npm)


mtho11:
It turns out the test failures were due to Angular 1.6 and not anything here. So #52 addresses this and needs to be merged first.

package.json
```diff
@@ -8,7 +8,8 @@
     "test": "karma start",
     "prepublish": "typings install && webpack && webpack --config webpack.vendor.config.js",
     "build-docs": "jsdoc -c jsdoc-conf.json",
-    "bower": "bower install"
+    "bower": "bower install",
+    "postinstall": "bower install"
```
himdel: From the [npm docs](https://docs.npmjs.com/misc/scripts): `install, postinstall: Run AFTER the package is installed.`

Meaning this change would make `npm install manageiq-ui-components` break in projects that don't use `bower`..

mtho11: Hey, I must have missed this over the Holidays. This is a good catch @himdel!
Changing to `preinstall`.

mtho11: @karelhala is the `bower` npm target  ("bower": "bower install") even used?
Can we get rid of it?

karelhala:
@mtho11 well we need it there because we still use patrernfly from bower. But I'm working on moving from bower to npm, once we'll there we no longer will need this task in `package.json`

himdel:
OK, there's a conflict but otherwise I think we can merge.. :)

mtho11:
Link to @karelhala upcoming [Bower Removal PR](https://github.com/ManageIQ/ui-components/pull/54) to watch.

mtho11:
**Consequences**: Just like _npmshrinkwrap_, when new versions of the library are published, the `yarn.lock` file needs to be updated(`yarn update`) and published as well.

himdel:
> Consequences: Just like npmshrinkwrap, when new versions of the library are published, the yarn.lock file needs to be updated(yarn update) and published as well.

Yeah, that sounds *wrong*. I don't think libraries should be restricting dependencies that way.

Can we skip adding `yarn.lock` to git?

himdel:
.. unless this doesn't affect the package that includes this library, and only affects the build.. Then no objections :)

mtho11:
IMO it actually sounds _right_. I have personally been bit too many times by indeterministic builds and transitive packages changing unexpectedly. But ultimately it is your decision, and I'm fine with whatever you decide.

Additional benefits are congruency with [manageiq-ui-service](https://github.com/ManageIQ/manageiq-ui-service) builds and yarns ultimate adoption by the Rails community for inclusion in Rails 5.1+.

mtho11:
If it is just the additional step to keep in sync we could just add an npm run script to do our own custom publish or something like that?

This is my first time using yarn so other yarn best practices probably exist.
@AllenBW any ideas? ^^

AllenBW:
HI @himdel @mtho11 !!!!! 👋 

Is the question general best practices?  If so I would like to say these few random thoughts
* always use `yarn upgrade-interactive`
* lock your yarn version for travis (in the past, versions have caused much 💔 )
* gotta save the yarn.lock file to the repo, otherwise it defeats the point of using it 🌮 

If I missed the question... 🙀

mtho11:
Thanks @AllenBW for your quick responses. 🥇

himdel:
OK, maybe I'm thinking about this wrong.

How does having `yarn.lock` in ui-components affect projects that include ui-components as a dependency?

I am 100% behind using yarn for manageiq-ui-*, as these are the end product, it definitely makes sense to fix the dependencies there.

ui-components on the other hand is a library. So.. how does this affect things?

If the end result is that `dist/ui-components.js` will get built using the same tools consistently, that's definitely a good thing :).

But if the end result is that this imposes additional version constraints on manageiq-ui-*, for every package mentioned here, that's the bad thing I'm afraid of. 

So that's what I'm asking, would it? (I originally thought yes, now I'm leaning towards no because calling `yarn` in miq should *not* read this repo's yarn.lock, should it?)

AllenBW:
So [here's an example of a dep (sui uses) with a yarn.lock published ](https://github.com/angular-ui/ui-router). 
You'll notice though, when install, the file in question is not present. 
screen shot 2017-01-12 at 7 49 56 am

Necessary for development, not for deployment.

himdel:
In that case, forget what I said, no objections from me :).

Should I click the button or do we have some blockers?

mtho11:
@himdel Let's click the button then!

PR #54, opened 2017-01-10
Remove bower

karelhala:
Bower dependencies are no longer needed. So bower was removed from dev dependencies and bower packages were moved to npm. 

This pull request removes lot's of packages, so same as with https://github.com/ManageIQ/ui-components/pull/41 
1. Remove `node_modules` and `lib` folder
2. Run `npm install`. 

For now vendor is not watched while development, so whenever you need to add new library you have to re-run `npm run install-vendor`. Vendor file is too large, it can probably be shrinked by removing a lot of libraries, which are not needed for our development, but I followed ng-patternfly's guide and used most of these libraries to make it work.

#### How to load external library
If you want to load external library, for example some angularJs module, you have to do these two steps:
1. Install it via npm: `npm install --save-dev ${LIB_NAME}`
2. Add it to `src/vendor.ts` (this will load it when building vendor part of application)

### How to load external window library
When new external window library is needed, for example we'd like to use [ramda.js](http://ramdajs.com/), we have to expose it from vendor file, otherwise it will be loaded twice (in vendor and in application - application after build will be much larger).
1. Install it via npm: `npm install --save-dev ramda`
2. Add it to `src/vendor.ts` (this will load it when building vendor part of application)
3. Add it to `webpack.vendor.config.js` to `ProvidePlugin`
The `key` under which will be this library available (`_` for lodash, `$` for jQuery ....), `value` is name of this library. So if we stick with Ramda.js as example new row `R: ramda` would be added and `R` command is available in window.
4. [Optional] If you need to use this library in application you have to load it from global space. Add it to `webpack.config.js` to field `externals`. Similar logic  as with `vendor` part. However `key` is value which will be used when importing this library and `value` is window's object. So if we stick with Ramda as example, and if it was exposed as `R` in step 3. we'll load it by adding new line `ramda: R`.


himdel:
@karelhala there's a conflict here..

karelhala:
@himdel all done.

package.json
```diff
@@ -6,12 +6,12 @@
   "scripts": {
     "start": "webpack --watch",
     "test": "karma start",
-    "prepublish": "webpack",
-    "install-vendor": "npm run bower && webpack --config webpack.vendor.config.js",
-    "build": "bower install && webpack --config webpack.vendor.config.js && webpack",
-    "build-docs": "jsdoc -c jsdoc-conf.json",
     "bower": "bower install",
-    "preinstall": "bower install"
+    "preinstall": "bower install",
```
himdel: `preinstall` and `bower` are still needed here?

karelhala: Good catch, they are not needed. Removed and repushed.

package.json
```diff
@@ -6,12 +6,10 @@
...
     "start": "webpack --watch",
     "test": "karma start",
-    "prepublish": "webpack",
-    "install-vendor": "npm run bower && webpack --config webpack.vendor.config.js",
-    "build": "bower install && webpack --config webpack.vendor.config.js && webpack",
-    "build-docs": "jsdoc -c jsdoc-conf.json",
-    "bower": "bower install",
-    "preinstall": "bower install"
+    "build": "webpack --config webpack.vendor.config.js && webpack",
+    "prepublish": "webpack && npm run install-vendor",
```
himdel: Guessing bad rebase.. should be `prepublish: "webpack",`, right?

karelhala: Yeah... I guess Friday 13th has a lot of impact on me :smile:

himdel: Friday friday... :notes:

himdel:
So...

```
$ ag bower
tsconfig.json
16:    "bower"

application-settings.js
9:  bowerLibs: 'libs/',

README.md
35:* Bower (soon to be removed and replaced with npm/yarn)
58:yarn global add typings bower webpack wiredep-cli typescript typescript-formatter

webpack.config.js
80:      {test: /\.html$/, loader: 'raw', exclude: /(node_modules|libs|dist|tsd|bower)/},

demo/template-index.ejs
48:<!-- bower javascripts -->

webpack.vendor.config.js
31:      {test: /\.html$/, loader: 'raw', exclude: /(node_modules|libs|dist|tsd|bower)/},
```

I'm guessing those webpack config mentions (and tsconfig maybe?) are valid, but the others should probably go as well?

karelhala:
@himdel you are right. I kept them there, because I was lazy to filter readme and such. But now there should be no occurrence of bower :smile:

PR #55, opened 2017-01-10
Angular components for the Dialog Editor

romanblanco:
https://github.com/ManageIQ/manageiq-ui-service/pull/54


src/dialog-editor/components/modal-field-template/modal-field-template.js
```diff
@@ -5,7 +5,7 @@
     .directive('dialogEditorModalFieldTemplate', function() {
       return {
         templateUrl: function(tElement, tAttrs) {
-          return 'app/components/dialog-editor-modal-field-template/' + tAttrs.template;
+          return tAttrs.template;
```
romanblanco: Still not sure about this one, have to try if it works before removing WIP

karelhala:
The build is failing, because you are mixing `.js` files in typescript. You should rewrite any javascript file to typescript.

mtho11:
@karelhala you beat me to it ;)

Steps:
- [ ] Convert all *.js to *.ts
- [ ] Move from common.js modules to ES6 modules
- [ ] Move syntax from ES5 to ES6/typescript

himdel:
@karelhala, @mtho11 I don't agree. Is there any technical reason ui-components needs to be typescript-only?

It makes sense to me to put all our shared ui components here, whether they're written in ES5, ES6 or TypeScript.

(I'm not saying this component shouldn't be typescript, if it is, even better .. but I am saying that it should not be a requirement.)

karelhala:
Well the biggest problem is that we don't have loader for `.js` files, so they are not compiled from es6 to es5 (which is reason why this particular build is failing). However this can be solved by adding babel loader alongside with eslint rules.

But does it makes sense to mix JS and TS in one project? I'd say more reasonable approach would be not to use all TS's features, but still rewriting every JS file to typescript (you don't have to create interfaces, use types or whatever, just replace IIF with `export class` and leave the rest as is). To be honest, I'd prefer going either full babel with ES6, or Typescript. But mixing these two together feels, like we could not decide what to use so we used both of them.

I understand that it's easier to take already functioning component and just use it here, it sure is faster than rewriting whole thing over and over again. So what about we put babel loader to webpack and allow using JS files in build process (whether they are written in ES6 or 5) and when such component is introduced we create new issue which states that and when somebody has a bit of spare time he will rewrite such component to typescript. So we have more consistent code base.

mtho11:
What about **not** adding the babel loader and just keeping everything *.ts and just replacing the IFFE blocks with `export class` like @karelhala says above. The typescript compiler should handle both ES5 and ES6 syntax, but this way we are not complicating the webpack build and not totally opening the gate to ES5 whatever.  I guess this is my endorsement for what @karelhala  said above 🥇

himdel:
Well, I guess the difference is that I'm thinking of this as a bag for reusable components, not a library with a consistent style and language.. So in my mind, we should support all the options.

That said, @romanblanco seems OK with updating the PR to something typescript-like, so I guess we don't have to have that discussion here and now.

But just to be clear about this, TS is *not* able to handle ES6 syntax. `let foo = {}; foo.bar = 5;` .. see if it works :).

mtho11:
@himdel It does if you play with implicit **any** stuff. There are compiler settings for this. But you can also just do: `let foo:any = {}; foo.bar = 5;` and that works.

I just tried it in: http://www.typescriptlang.org/play/

karelhala:
@romanblanco you do not need to add 3rd party libraries to your `index.ts` via `require` just make sure they will be loaded in application which will use your module. If you'd were to require them, they'd become part of our library and we'd have to take care of them (hence the drop in coveralls).

romanblanco:
@karelhala Thanks, I've just removed them :)

package.json
```diff
@@ -21,6 +21,9 @@
     "node": ">= 6.0.0",
     "npm": ">= 3.8.1"
   },
+  "files": [
+    "dist/js/ui-components.js"
```
himdel: I guess you don't want this .. if I'm reading the [docs](https://docs.npmjs.com/files/package.json#files) correctly, this would limit the npm package to just that one file

src/dialog-editor/services/modal/modal.ts
```diff
@@ -0,0 +1,298 @@
...
+            this.DialogEditor.activeTab
+          ].dialog_groups[
+            this.dialog.boxId
+          ].dialog_fields[
+            this.dialog.fieldId
+          ]
+        );
+        // load categories from API, if the field is Tag Control
+        if (this.modalData.type === 'DialogFieldTagControl') {
+          this.resolveCategories(this.CollectionsApi).then(function(categories: any) {
```
himdel: `this.resolveCategories()` - resolveCategories doesn't seem to take any arguments

romanblanco:
@dtaylor113 This PR is now ready, you can do your changes based on this

src/dialog-editor/components/modal-field-template/modalFieldTemplateComponent.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @name dialogEditorModalFieldTemplate
+ * @description
+ *    Component contains templates for the modal for each field type
+ * @example
+ * <dialog-editor-modal-field-template ng-switch-when="DialogFieldTextBox"
+ *                                     template="text-box.html">
+ * </dialog-editor-modal-field-template>
+ */
+export default class ModalFieldTemplate {
+  public template = ['$element', '$attrs', function($element: any, $attrs: any) {
```
himdel:
```
// @ngInject
public template = ['$element', '$attrs', function($element: any, $attrs: any) {
```

mtho11:
Is this WIP or complete?

romanblanco:
@mtho11 Complete, waiting for one more review

src/dialog-editor/components/tab-list/tab-list.scss
```diff
@@ -0,0 +1,13 @@
+.delete-tab {
```
himdel: These classes are not in the generated css .. missed it? :)

src/dialog-editor/components/modal-field-template/modalFieldTemplateComponent.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @example
+ * <dialog-editor-modal-field-template ng-switch-when="DialogFieldTextBox"
+ *                                     template="text-box.html"
+ *                                     modal-data="vm.modalData">
+ * </dialog-editor-modal-field-template>
+ */
+export default class ModalFieldTemplate {
+  /*@ngInject*/
+  public template = ($element: any, $attrs: any) => {
+    return require('./' + $attrs.template);
```
karelhala: can be simplified:
```typescript
public template = ($element: any, $attrs: any) => require(`./${$attrs.template}`);
```

src/dialog-editor/components/modal-field-template/modalFieldTemplateComponent.ts
```diff
@@ -0,0 +1,40 @@
...
+  };
+  public scope: boolean = true;
+  public controllerAs: string = "vm";
+  public bindings: any = {
+    modalData: '=',
+    addEntry: '=?',
```
karelhala: Good one on the optional parameter. Howerver, based on the `=` I thought they are two way databinding, is there a reason not to use callback?
```javascript
public bindings: any = {
  modalData: '=',
  addEntry: '&?',
  removeEntry: '&?',
  resolveCategories: '=?',
  categories: '=?'
}
```

src/dialog-editor/components/tab-list/tabListComponent.ts
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Add a new tab to the list.
+   * New tab is automatically appended to the last position of the list and
+   * set as active.
+   * @memberof TabListController
+   * @function addTab
+   */
+  public addTab() {
+    // deactivate currently active tab
+    this.tabList.forEach(function(tab) {
```
karelhala: No need to use `function`
```typescript
this.tabList.forEach(tab => tab.active = false);
```

src/dialog-editor/components/tab-list/tabListComponent.ts
```diff
@@ -0,0 +1,163 @@
...
+        this.tabList[id - 1].active = true;
+      } else if ((this.tabList.length - 1) > this.tabList[id].position) {
+        // active tab was not at the end → new active tab is on following index
+        this.tabList[id + 1].active = true;
+      }
+    }
+    // remove tab with matching id
+    _.remove(
+      this.tabList,
+      function(tab) {
```
karelhala: No need to use `function`
```typescript
_.remove(this.tabList, tab => tab.position === id);
```

src/dialog-editor/components/toolbox/toolboxComponent.ts
```diff
@@ -0,0 +1,132 @@
...
+  public label: string;
+  public placeholders: any;
+
+  constructor(type: string,
+              icon: string,
+              label: string,
+              options: any = {}) {
+    this.icon = icon;
+    this.label = label;
+    this.placeholders = Object.assign({
```
karelhala: I think [object-rest-spread](https://github.com/sebmarkbage/ecmascript-rest-spread) looks better, than `Object.assign` but that's on preferences.
```typescript
this.placeholders = {
      name: '',
      description: '',
      type: type,
      display: 'edit',
      display_method_options: {},
      read_only: false,
      required: false,
      required_method_options: {},
      default_value: '',
      values_method_options: {},
      label: label,
      position: 0,
      dynamic: false,
      show_refresh_button: false,
      load_values_on_init: false,
      auto_refresh: false,
      trigger_auto_refresh: false,
      reconfigurable: false,
      visible: true,
      options: {
        protected: false,
      },
      resource_action: {resource_type: 'DialogField', ae_attributes: {}},
      ...options
}
```

src/dialog-editor/services/modal/modalService.ts
```diff
@@ -0,0 +1,295 @@
...
+   */
+  public showModal(tab: number, box: number, field: number) {
+    let modalOptions = {
+      template: require('./modal.html'),
+      controller: ModalController,
+      controllerAs: 'vm',
+      size: 'lg',
+      resolve: {
+        dialogDetails: function() {
+          return {tabId: tab, boxId: box, fieldId: field};
```
karelhala: No need to use `function`
```Typescript
dialogDetails: () => {
  return {tabId: tab, boxId: box, fieldId: field};
}
```

src/dialog-editor/components/toolbox/toolboxComponent.ts
mtho11: I agree, but these are trivial things. But ultimately need to migrate their way to tslint or whatever. If we want them to become idiomatic for MiQ.

webpack.config.js
```diff
@@ -89,6 +89,7 @@ module.exports = {
   plugins: plugins,
   externals: {
     'angular': 'angular',
-    'lodash': '_'
+    'lodash': '_',
+    '__': '__'
```
mtho11: Nice Addition!!

src/dialog-editor/components/box/boxComponent.ts
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Remove box and all its content from the dialog.
+   * @memberof BoxController
+   * @function removeBox
+   * @param {number} id as index of removed box
+   */
+  public removeBox(id: number) {
+    _.remove(
+      this.dialogTabs[this.DialogEditor.activeTab].dialog_groups,
+      function(box) {
```
romanblanco: Thanks, I've used the arrow function so far on places, where I had to pass `this` (http://stackoverflow.com/questions/16157839/typescript-this-inside-a-class-method helped). 

I'll replace also the other functions, you've mentioned 👍

karelhala:
@romanblanco found the problem why build started to fail. Looks like Webpack 1 ignored any error and exits every time with 0, even tho some build errors occurred. However webpack 2 is coded better (for us, worse) and does not ignore build errors. However we can force him to ignore these errors, just go to `package.json` and add `webpack || true` to prepublish. It's an ugly sollution but it works and we can merge this PR and work on webpack/typescript settings in seperate PR.
```json
...
 "scripts": {
    "start": "webpack --watch",
    "test": "karma start",
    "prepublish": "webpack || true",
    "build": "webpack --config webpack.vendor.config.js && webpack -p",
    "install-vendor": "webpack --config webpack.vendor.config.js",
    "build-docs": "jsdoc -c jsdoc-conf.json"
 },
...
```

romanblanco:
Thank you, @karelhala!

himdel:
Merging, will release a new version together with #58.

The problem with not being able to see the DialogEditor service has something to do with yarn and branches, instaliing via `npm` worked just fine.

mtho11:
@himdel What exactly is the issue with yarn and branches?

himdel:
@mtho11 no idea about the exact issue, but have a read in  https://github.com/ManageIQ/manageiq-ui-service/issues/478 :).

As it is now (with `"manageiq-ui-components": "romanblanco/ui-components#dialog_editor"` in `package.json`, it would install a version where `dist/js/ui-components.js` did not include the substring `miqStaticAssets.dialogEditor` at all :).

himdel:
Released 0.0.12

PR #56, opened 2017-01-11
Change prepublish not to use npm run

karelhala:
As @martinpovolny and @himdel stated there is issue with when script in `packag.json` contains `npm run whatever`, so instead using placeholder tasks in `scripts` field use their contents.

After some discussion with @himdel much cleaner solution is to have as simple `postinstall` as possible, so now it only packs this library via `webpack` and no vendors are generated. To generate vendor files you have to run `npm run build` (installs bower dependencies + packs vendor files together + builds this library) or `npm run install-vendor` (installs bower dependencies +  packs vendor files together)


himdel:
Thanks @karelhala! 

One question, I noticed this changes the order of `webpack --config webpack.vendor.config.js` and `webpack`, tried to see if it matters or not .. and it seems like the `vendor` bit doesn't actually do anything. Does it?

Or rather, it touches `dist/js/vendor.js` which is gitignored and doesn't end up in the final package.

Do we actually need this? Or, assuming it's only for the demo, shouldn't all that demo related stuff live there?

karelhala:
Well `vendor` is just for demo and test purposes, so thinking about it and as you pointed out it does not makes much of a sense to have it inside `postinstall`, so I think moving this to `build` script might be a better option.

However to have `vendor` inside dist is convenient for demo and test purposes, so I think we can keep it there, I'll just move it from `postinstall`.

PR #57, opened 2017-01-23
Webpack update

karelhala:
Since webpack 2 left beta stage and is now in stable `2.2.0` [roadmap](https://webpack.github.io/docs/roadmap.html) we can migrate to it and lower final file size.

### Webpack 2 changes
After minification the result file dropped from **~90KB** to **31.2KB** Not much of a change, but after we'll add new components the difference will be much bigger.
 * Used full name of loader (we can't omit `-loader`).
 * Per loaders is deprecated, used `enforce: 'pre'` instead.
 * Replaced `loaders` with `rules`.
 * Empty `resolve.extension` is forbidden.
 * `ExtractTextPlugin` requires one param with `loader` and `fallbackLoader` defined.
 *  Updated `ExtractTextPlugin`
 * Running `yarn run build` now runs `webpack -p` which will minify and optimize final project.

### Vendor changes
Since we moved from bower provided patternfly to npm's we no longer need all dependencies to be included in `vendor.ts` file for testing. Dropped from **~9.2MB** to **~3MB** this is still not ideal, but given that we are using unminified versions of angular, and patternfly in test (demo application) we can settle for now.
 * Removed unnecessary patternfly's dependencies.


himdel:
Agreed completely, tested that manageiq's toolbars work after this change.. so.. doing the thing :).

PR #58, opened 2017-01-30
Gtl navigation

karelhala:
As @dclarizio proposed after checking tile view navigation for both tiles and table needs to be better aligned, the problem is that pagination did not took enough space and bottom toolbar is empty.

### Changes
* Rearrange items in navigation
  * Select all | **Per page** | Sort By | Pagination -> Select all | Sort By | **Per page** | Pagination
* Move `per page` closer to `pagination`
* Use more space and float `Per page` and `Pagination` items to right.

#### Table
selection_120
#### Tile
selection_121

#### Table
selection_116
#### Tile
selection_115

#### Table
selection_123
#### Tile
selection_122

#### Table
selection_118
#### Tile
selection_119


himdel:
Visually the change is in good direction, resolving the conflict (in the `.map` file - no point waiting since this will get overwritten when releasing a new version anyway) and merging.

himdel:
Released 0.0.12

PR #59, opened 2017-02-07
Add dependencies for dialog component to vendor, add total of to paging

karelhala:
As @himdel proposed it would be nice to have total of records and show number of visible items.

### Behavior
* Current page is last -> Show `startIndex` - `allItemsCount`
* Current page is **not** last -> Show `startIndex` - `endIndex`

Start index is calculated like `(currentPage - 1) * perPage`
End index is calculated like `currentPage - * perPage`

#### Before
selection_129

#### After
selection_128


karelhala:
During this PR I also fixed failing of demo application via adding dependencies to `vendor.ts` and including dependencies for dialog editor in demo app.

src/gtl/components/data-table/data-table.html
```diff
@@ -23,8 +23,14 @@
         </div>
       </div>
     </div>
+    <div class="col-md-1 col-lg-1 col-xs-1 miq-page-counter">
+      <label>{{tableCtrl.settings.startIndex + 1}} - {{tableCtrl.settings.endIndex + 1}}
+        {{tableCtrl.MiQTranslateService.translateString('of')}}
```
himdel: I think for i18n, this would have to be more like "%d - %d of %d" .. otherwise, there's too many ways of translating "of" that don't make sense.

himdel:
Is the translate service code something that's just a placeholder for now, or is it supposed to work already somehow?

karelhala:
That's just placeholder, so when `__` is present in window, it will use this method, otherwise just return string which was passed for translation.

himdel:
OK, so.. I guess I can merge, and we can figure out how to make it work later, right? :)

src/gtl/components/data-table/data-table.html
```diff
@@ -24,11 +24,7 @@
       </div>
     </div>
     <div class="col-md-1 col-lg-1 col-xs-1 miq-page-counter">
-      <label>{{tableCtrl.translateString(
-        (tableCtrl.settings.startIndex + 1) +
-        ' - ' + (tableCtrl.settings.endIndex + 1) +
-        ' of ' + tableCtrl.settings.items)}}
-      </label>
+      <label>{{tableCtrl.settings.startIndex}} - {{tableCtrl.settings.endIndex}} : {{tableCtrl.settings.items}}s</label>
```
himdel: Sorry.. trailing `s` ;) ..and the same in tile view

karelhala: Good notice, my `ctrl` does not work really great so hitting `ctlr` + `s` usually ends up in just random `s` in code :smile:

karelhala:
@himdel after discussion about translating total records of I came up with this... if `settings` object contains function `translateOf(start, end, total)` call it, if not return `start - end of total`.

Implementation of `translateOf(start, end, total)` can be:
```Javascript
this.settings.translateOf = function(start, end, total) {
  return __('%d - %d of %d', start, end, total);
}
```

himdel:
Sounds reasonable :)

except reading the code seems to indicate that it has to be `this.settings.translateTotalOf` ;)

himdel:
Released `v0.0.13`

PR #61, opened 2017-02-16
No select box in nested views

karelhala:
### Removed selectbox and select all in nested lists

When report is shown in nested screens do not allow user to select items, either via clicking on select input, or on clicking on item, or by clicking on select all. (There is still hover change of tile however)
peek 2017-02-16 13-41

### Clicking anywhere on tile will select it
Before
peek 2017-02-16 13-48
After
peek 2017-02-16 13-49

### Send selected records in `report_data`
When some records are selected, we can now send `report_data` GET with these selected IDs.
So now the full payload can look like:
```
model:someModel
active_tree:widgets_tree
explorer:true
records[]:recordID1
records[]:recordID2
page:1
ppsetting:5
sort_choice:1
is_ascending:true
model_id:someModelId
```


karelhala:
ping @himdel @mtho11

himdel:
Code looks good, but not sure where to test in the UI (mostly because of the existing issues in ManageIQ/manageiq-ui-classic#26)

PR #66, opened 2017-05-03
Enable map for main css file in bower distribution

karelhala:
Since `ActionController` complains that `/assets/manageiq-ui-components/dist/css/ui-components.css.map` is missing we should ship map file for styles.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -4,6 +4,7 @@
   "description": "UI components for ManageIQ project",
   "main": [
     "dist/css/ui-components.css",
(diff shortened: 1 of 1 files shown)


martinpovolny:
Fixes: 
```
ActionController::RoutingError (No route matches [GET] "/assets/manageiq-ui-components/dist/css/ui-components.css.map"):...
```

jzigmund:
I can still see the issue, reported to @karelhala

PR #68, opened 2017-05-04
Remove mapped files from gitignire

karelhala:
As @jzigmund mentioned in https://github.com/ManageIQ/ui-components/pull/66#issuecomment-299181036 `dist/css/ui-components.css.map` was still not present in distributed bower due to policy at Bower, that only one file with can be treated as main with given filetype [main-list](https://github.com/bower/spec/blob/master/json.md#user-content-main). So we do not ignore `.map` files in dist folder for github and also do not ignore whole folders `js` and `css` in `dist`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ vendor.css
 coverage
 
 !dist/js/ui-components.js.map
+!dist/css/ui-components.css.map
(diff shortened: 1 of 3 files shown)


himdel:
@jzigmund can you `bower install` and re-check please?

jzigmund:
@karelhala @himdel I can confirm that the issue has been fixed

PR #70, opened 2017-05-19
minify both CSS and JS, up version of TS

karelhala:
No changes to code except I increased version of typescript and rebuilt application using `npm run build` which builds the app using `-p` so the final application is minified.


himdel:
Cc @skateman this should fix the `lodash` problems you've encountered.

PR #71, opened 2017-05-22
Update documentation, add sextion about running build process

karelhala:
As more users will begin to use this repository we have to keep docs consistent, so update docs about run process, where you have to run `npm run bild` or `yarn run build` in order to generate JS dependencies.


himdel:
LGTM, though I think it's spelled "section" ;)

PR #74, opened 2017-05-25
Customize Dialog Editor components for use in Ops UI

romanblanco:
- `CollectionsApi` is missing in OpsUI -> using `API`
- typo fix, naming fix


src/dialog-editor/services/modal/modalService.ts
```diff
@@ -98,7 +98,9 @@ class ModalController {
       attributes: ['description', 'single_value', 'children'],
     };
```
martinpovolny: the query method is not available here?

if so then you don't need the `options` do you?

romanblanco: @martinpovolny thank you, updated.

martinpovolny:
Looks good to me.

@himdel, @karelhala, @mtho11 : please, merge this if it looks good to you.

PR #76, opened 2017-06-20
Pagination component as standalone

karelhala:
Refactor pagination component so it will be one component holding all settings of pagination.
* [x] Move component to separate files
* [x] Make component correspond to specific API (select All, change per page, change page, change sort (both sort by and sort column))
* [x] Add tests to this component.

The API is:

| API value  | Description |
| ------------- | ------------- |
|  settings  |  Basic settings of pagination (sort by, titles, translate function and such)  |
|  sortBy  |  Sort by settings (Might be merge this into settings object)  |
|  perPage  |  Per page settings (Might be merge this into settings object)  |
|  onSelectAll  |  Function which is fired when select all is clicked  |
|  onChangeSort  |  Function which is fired when sort by or sort direction is changed  |
|  onChangePage  |  Function which is fired when page is changed  |
|  onChangePerPage  |  Function which is fired when number of items per page is changed  |

### DIFF
--- a/demo/controllers/dataTableController.ts
+++ b/demo/controllers/dataTableController.ts
@@ -31,6 +31,9 @@ export default class DataTableController {
       this.cols = data.cols;
       this.settings = data.settings;
       let start = (this.settings.current - 1) * this.settings.perpage;
+      this.settings.translateTotalOf = (start, end, total) => {
+        return `${start + 1} - ${end + 1} of ${total}`
+      };
       this.onLoadNext(start, this.settings.perpage);
       this.onSort(2, true);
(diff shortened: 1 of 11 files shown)


martinpovolny:
@karelhala : in the context of pagination and GTL components where is the filtering or search state being stored?

I am trying to understand the whope picture with regard to https://github.com/ManageIQ/manageiq-ui-classic/pull/1553/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR392

himdel:
Tested in the UI, seems to work as before, merging :)

PR #77, opened 2017-06-22
Clean build process so there are no more errors with missing __ function

karelhala:
This PR fixes errors which were shown in console if any phase of build was ran (`npm run build` or `npm start`). Helper function was introduced, which checks if `__` exists on window object and if it does not exists it will return string which was passed to it, if it exists it will return result of `__(translateString)`.

No UI changes.

### DIFF
--- a/src/dialog-editor/components/toolbox/toolboxComponent.ts
+++ b/src/dialog-editor/components/toolbox/toolboxComponent.ts
@@ -10,7 +10,7 @@ class DialogField {
               options: any = {}) {
     this.icon = icon;
     this.label = label;
-    this.placeholders = (<any>Object).assign({
+    this.placeholders = Object.assign({
       name: '',
       description: '',
       type: type,
--- a/src/dialog-editor/components/toolbox/toolboxComponent.ts
+++ b/src/dialog-editor/components/toolbox/toolboxComponent.ts
@@ -1,3 +1,4 @@
+import {__} from '../../../common/translateFunction';
 class DialogField {
   public icon: string;
   public label: string;
@@ -9,7 +10,7 @@ class DialogField {
               options: any = {}) {
     this.icon = icon;
     this.label = label;
-    this.placeholders = Object.assign({
+    this.placeholders = (<any>Object).assign({
       name: '',
       description: '',
       type: type,
--- a/src/common/translateFunction.ts
+++ b/src/common/translateFunction.ts
@@ -1,0 +1,7 @@
+export function __(translateString) {
+  if (window.hasOwnProperty('__')) {
+    return window['__'](translateString);
+  } else {
+    return translateString;
+  }
+}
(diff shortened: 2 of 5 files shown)


karelhala:
@romanblanco I was unable to find where dialog editor is used in SUI so I am unable to test this directly in there. Could you try it if it does not break anything?

romanblanco:
@karelhala it is not used in SUI anymore. It was removed in PR https://github.com/ManageIQ/manageiq-ui-service/pull/688.

Right now, there is just PR adding it into Ops UI: https://github.com/ManageIQ/manageiq-ui-classic/pull/1398, otherwise it's not used anywhere right now

romanblanco:
@karelhala tested, works for me 👍 no more errors with missing `__` function 🎈 🎆

src/dialog-editor/components/toolbox/toolboxComponent.ts
```diff
@@ -9,7 +10,7 @@ class DialogField {
               options: any = {}) {
     this.icon = icon;
     this.label = label;
-    this.placeholders = Object.assign({
+    this.placeholders = (<any>Object).assign({
```
himdel: This one sounds wrong .. any way to teach tslint about es6-shim? Or, to assume all ES6 methods are available?

himdel: (the method *is* defined in `./node_modules/typescript/lib/lib.es6.d.ts`, just not sure how to make it use that without using `target: "es6"`)

himdel: ... actually, this works (but not sure if quite the right place..)

```diff
diff --git a/tsconfig.json b/tsconfig.json
index 3093662..c97da88 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
     "typeRoots" : ["./node_modules/@types"]
   },
   "files": [
+    "node_modules/typescript/lib/lib.es6.d.ts",
     "src/index.ts"
   ],
   "exclude": [
```

himdel: @karelhala So looks like this is the proper fix:

```diff
diff --git a/tsconfig.json b/tsconfig.json
index 3093662..4d3d5e9 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,8 @@
     "module": "commonjs",
     "removeComments": false,
     "sourceMap": true,
-    "typeRoots" : ["./node_modules/@types"]
+    "typeRoots" : ["./node_modules/@types"],
+    "lib": ["es2015"]
   },
   "files": [
     "src/index.ts"
```

karelhala: Thank you, looks much better! :+1:

himdel:
@karelhala I guess you need to rebase..

I'm starting to wonder whether we need a separate repo (autogenerated on change) for `dist/` so that we don't have to keep the built version here and have conflicts all the time.

mtho11:
@himdel that sounds like a cleaner solution having `dist/` autogenerated to another repo.

PR #78, opened 2017-06-29
Introducing the fonticon picker component

skateman:
A button-like element for selecting a fonticon from multiple font families. This initial implementation provides a hidden input field to store the selected icon for the compatibility with our non-angular custom button form. This part should probably refactored in the future after we converted the custom button form to angular.
screenshot from 2017-06-29 12-09-58
If you click on the right caret next to the fonticon, it displays a modal where you can select a fonticon from multiple font families:
screenshot from 2017-07-07 18-20-04

**Usage:**
```html
<miq-fonticon-picker input-name="foo" btn-class="bar" selected="fa fa-info">
  <miq-fonticon-family selector="fa" title="Font Awesome"></miq-fonticon-family>
  <miq-fonticon-family selector="glyphicon" title="Glyphicons"></miq-fonticon-family>
  ...
</miq-fonticon-picker>
```
Where `input-name` is being used as the name of the hidden field containing the selected icon's class. The `btn-class` is set as a class on both the button and the caret, defaults to `btn-default`. And the `selected` should contain the CSS class of the initially selected icon. Inside the `<miq-fonticon-picker>` you can define multiple `<miq-fonticon-family>` components, each describing a font family with its `selector`. The families are rendered as individual tabs with their `title` attribute as the tab label.

**TODO:**
- [x] Implemented the basic component
- [x] Internationalization of the modal title & buttons
- [x] Styling of the icons in the modal

cc @epwinchell, @karelhala, @himdel


himdel:
LGTM, except for the styling bits :). (Waiting for those before merging :).)

martinpovolny:
Nice! :+1:

martinpovolny:
For future PRs it would be beneficial if you could put the changes to the generated files to a separate commit.

skateman:
@martinpovolny or we could simply put these files into `.gitignore` and ask Travis to do the build and force add these files after each merge.

src/fonticon-picker/services/fonticonService.ts
```diff
@@ -0,0 +1,45 @@
+import * as _ from 'lodash';
+
+export default class FonticonService {
+  public fetch(families) {
```
karelhala: This function feels too complicated and too big. I am not big fan of regular expressions, hard to read in couple of hours after writing them.

What do you think about functional programming? And pure functions? Plus would be nice to provide some interfaces (but that is nice to have and not necessary)
```Typescript
import * as _ from 'lodash';

export default class FonticonService {
  public fetch(families: any): any {
    return _.reduce(families, (result: any, value: any) => {
      result[value.selector] = FonticonService.filterRules(value.selector);
      return result;
    }, {});
  }

  private static filterRules(family: string): any[] {
    return _.chain(document.styleSheets)
      .map((oneSheet: any) => oneSheet.rules)
      .map((rule: any) => _.filter(rule, value => FonticonService.isFontIcon(value, family)))
      .filter((rules: any) => rules.length !== 0)
      .map((rules: any[]) => _.map(rules, (value: any) => FonticonService.clearRule(value.selectorText)))
      .flatten()
      .reduce((result: any[], value: string) => [...result, FonticonService.makeRuleObject(family, value)], [])
      .value();
  }

  private static isFontIcon(value, family): boolean {
    return value.selectorText && value.selectorText.indexOf(family) === 1 && value.cssText.indexOf('content:') !== -1;
  }

  private static clearRule(rule: string): string {
    return rule.replace('::before', '').split(',')[0].replace('.', '');
  }

  private static makeRuleObject(family, value): any {
    return {
      'class': `${family} ${value}`,
      'selector': `.${family}.${value}`
    }
  }
}
```

epwinchell:
@skateman Let's change the active color to "#0088ce" - (our standard blue active color rather than red.)

screen shot 2017-07-05 at 2 42 37 pm

src/fonticon-picker/services/fonticonService.ts
skateman: Yay, thanks!

skateman:
@epwinchell I'm pretty sure that the coloring won't work on patternfly fonticons as some of them are already colorized. Can we have a border/box shadow instead?

epwinchell:
Looks good.

skateman:
@karelhala :shipit:

martinpovolny:
> @martinpovolny or we could simply put these files into .gitignore and ask Travis to do the build and force add these files after each merge.

Any reason not to do that? @skateman, @karelhala, @himdel, @mtho11 ?

It will clean our PRs, and repo history.

skateman:
@martinpovolny definitely not in this PR, and for now it wouldn't help much. Because of bower, we need the binaries in the master and even if having them in `.gitignore`, force adding a file overrides the ignores and it keeps tracking the modified binaries locally :disappointed:

skateman:
After we fully migrate to `yarn/webpacker` in MIQ, we can use this for automated deployment: https://docs.travis-ci.com/user/deployment/npm/

martinpovolny:
> @martinpovolny definitely not in this PR, and for now it wouldn't help much. Because of bower, we need the binaries in the master and even if having them in .gitignore, force adding a file overrides the ignores and it keeps tracking the modified binaries locally :disappointed:

So that brings me back to my original request: 

**For future PRs it would be beneficial if you could put the changes to the generated files to a separate commit.**

skateman:
@martinpovolny adjusted, @karelhala ready to merge!

himdel:
Re the `dist/` discussion .. I was thinking we should have a separate repo for that, but the idea was mostly the same - created https://github.com/ManageIQ/ui-components/issues/81 to keep it in one place.

Loicavenel:
@skateman very good job.

Loicavenel:
@skateman Do you know how many icons do we have now?

skateman:
@Loicavenel *811* icons with *16 777 216* different colors so *13 606 322 176*, with words *thirteen billion, six hundred six million, three hundred twenty-two thousand, one hundred seventy-six* :wink:

PR #83, opened 2017-07-12
Automated build of binaries from master to a dev tag using Travis

skateman:
It resolves #81, but still needs an `username:token` pair for **miq-bot** to set as `GITHUB_AUTH` in Travis' secret environment variables. Basically something similar (or even the same) as we have for **font-fabulous**. @Fryguy ??

PR #86, opened 2017-07-13
Add missing semicolon in the fonticon picker

skateman:


himdel:
:+1: A good way to test #83. Merging when green..

PR #87, opened 2017-07-13
Fix bower deployment if the dist folder already exists

skateman:


himdel:
.. the idea being that we drop any old `dist/` before replacing it completely with the freshly built one.

And `mv -f tmp dist` will always rename, not move to `dist/tmp` as it did when `dist` already existed.

Waiting for green :+1:

ci/bower-dev.sh
```diff
@@ -7,6 +7,7 @@ mv bower.json bower.json.tmp
 git remote add upstream https://github.com/ManageIQ/ui-components.git
 git fetch upstream
 git checkout bower-dev
+rm -rf dist
```
Fryguy: This will delete also any ancillary files like the README and LICENSE, which might have side effects?

skateman: README and LICENSE are not in the `dist` folder

Fryguy: Oh I see...they are at the root of the repo.  :+1:

PR #89, opened 2017-07-20
Fix for issue with dropdown rendering

chalettu:
This is a fix for the siteswitcher not rendering the dropdown properly.


chalettu:
fix  Here is a demo of the component working as expected after the fix.

PR #90, opened 2017-07-21
NoEmitOnError flag for typescript compiler.

mtho11:
In an effort to not allow bad stuff into the build this flag will not compile the bad stuff. I'm seeing on other typescript projects that things get committed even though they raise compile errors (this should not be allowed IMO). Keeps it clean from the start.

wdyt?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,7 @@
     "target": "es5",
     "module": "commonjs",
     "removeComments": false,
+    "noEmitOnError": true,
     "sourceMap": true,
     "typeRoots" : ["./node_modules/@types"],
     "lib": ["es2015", "DOM", "ScriptHost"]


himdel:
I don't know.. on the one hand, this sounds very reasonable, on the other, we've had a bunch of errors for a loong time, and it never stopped the code from working.

I guess what I'm saying is that while I agree that compiler errors should stop the compiler from producing code, I also think that TS is too ambitious in what it calls errors.

To me, having complete typing info is a would-be-nice, not having it is at best a warning. Yet TS bails on that.

On the other *other* hand, if we can have 0 warnings and errors, and if this stops Travis from passing on PRs which add errors, any new breakages would probably come only from new versions, and we can handle that..

I guess what I'm saying is I'm more for than against :) Karel? :)

mtho11:
I guess for me allowing committing of errors (even if they are not really bad errors) is a bit sloppy. Sooner or later, one of those errors is going to be something bad.

karelhala:
It depends on what size of this project do we want. If it would be small project with couple of components and couple of developers, then I am ok with commiting errors.

BUT if more developers would come and the project would get larger we have to force them to follow simple principles and do not commit any error, since it can break the whole package if such error would happen.

In my personal opinion I think we should merge this and do not allow any errors. At least anybody who'd come to this project will see that there are no errors in build process.

martinpovolny:
I agree with @karelhala. In my view the whole point of Typescript is to enforce strict rules so that the projects and teams can scale. As we discusses on Talk: the "opinionated yes/no" -- this is similar.

With small team of "rock starts" you can say: "we ignore this but as it's really not a bug or it does not matter..." but with bigger team, this just does not fly.

himdel:
The consensus is clear, merged :)

We should keep an eye on travis, to make sure the specs actually fail if somebody adds an error.

PR #93, opened 2017-07-25
Fixed resolving entries

romanblanco:
fixed issue described by @d-m-u in https://www.pivotaltracker.com/story/show/144481043/comments/177154253

There was a change in API. We're not using numbers for id's anymore, but strings.

Thanks for help, @himdel!

screencast from 2017-07-25 12-08-32

/cc @martinpovolny


martinpovolny:
What about the change to src/dialog-editor/components/modal-field-template/modalFieldTemplateComponent.ts? Does not seem to be related to the `Number` change.

Is that correct?

romanblanco:
@martinpovolny That's the same issue as in https://github.com/ManageIQ/ui-components/pull/92

PR #96, opened 2017-07-26
Renamed package and bumped version number

chalettu:
Renamed package to scoped NPM package name and bumped it to next logical version number


chalettu:
@himdel this is the PR we discussed about bumping package number

chriskacerguis:
@himdel this look good to you?

PR #102, opened 2017-07-27
Create Unit tests for dialog-user component

chalettu:
@miq-bot add_label test

### DIFF
--- a/demo/data/dialog-data-refresh.json
+++ b/demo/data/dialog-data-refresh.json
@@ -1,0 +1,313 @@
+{
+"resources": [
+    {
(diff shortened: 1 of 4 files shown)


chalettu:
PV ticket https://www.pivotaltracker.com/story/show/148796889

chriskacerguis:
@chalettu can you please take a look, the code coverage dropped despite adding tests /shrug

chalettu:
I see lots of potential issues with how testing is currently being done.  The current process builds a single js file that the test runner then goes against to report on issues.  The problem is that doesn't allow us to report on test coverage for individual files.  The second issue and the issue that caused coverage to drop is due to the fact that coveralls seems to be including the actual unit test JS files themselves in its is score of the repo.  Those files need to be excluded.

PR #105, opened 2017-08-01
Add function binding for fonticon picker onchange handling

skateman:
For being able to use `miq-observe` with non-angular forms. @karelhala


karelhala:
LGTM

PR #107, opened 2017-08-02
Add font fabulous to build application, change service to work with all fonts families including font fabulous

karelhala:
### Adds
* New font family called font fabulous in demo app
* Correct display of new font family in demo app

selection_186


karelhala:
@skateman ping

PR #108, opened 2017-08-03
Use cssRules instead of rules which is supported in all browsers

karelhala:
Use of correct fetch of cssRules as suggested by @skateman and found by @AparnaKarve.

### DIFF
--- a/src/fonticon-picker/services/fonticonService.ts
+++ b/src/fonticon-picker/services/fonticonService.ts
@@ -24,8 +24,8 @@ export default class FonticonService {
   }
 
   private static clearRule(rule: string, family: string): string {
-    let splitRule = rule.split(family);
-    return family + splitRule[splitRule.length - 1].replace('::before', '').split(',')[0].replace('.', '');
+    let cleared = FonticonService.removeChars(rule.replace('::before', ''), '.');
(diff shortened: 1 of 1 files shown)


epwinchell:
@karelhala @skateman Tested in IE11. Everything looks good, except the FF tab:
screen shot 2017-08-03 at 11 33 52 am

karelhala:
@epwinchell looks like IE is not cooperating. However I fixed the issue with not correctly added icons with fonts, turns out that IE is trying to make me mad and throws `CSS3111: @font-face encountered unknown error.` Which looks like some kind of basic error for fonts. If you however check the HTML code icons are okay, you can select them and work with them, but you can't see them in demo environment... 
selection_189

skateman:
@karelhala LOL, I see I have to IE11 as well :wink:

karelhala:
Testing this in MiQ environment works fine, so I guess it's a problem with loading font-face from base64 blob (which should work, but IE refuses to work with it), or multiple URL sources. Anyway, the problem is only on IE 11 in test environment, so this should not be problem. 

selection_190

skateman:
@himdel this is good to merge, @h-kataria this should fix your issue in Firefox.

PR #111, opened 2017-08-07
Serve data jsons over POST method

karelhala:
After merging https://github.com/ManageIQ/ui-components/pull/110 data for demo application were not present because the request was using POST method for static data. This PR adds middleware option to browsersync where any request to `/data` will be changed to use GET method.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -26,7 +26,16 @@ const webpack = require('webpack'),
     new BrowserSyncPlugin({
       host: 'localhost',
       port: 4000,
-      server: {baseDir: [__dirname + settings.distFolder]}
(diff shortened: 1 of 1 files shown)


karelhala:
@himdel can you please look at it?

himdel:
We're not using any GET params in the demo, right? :) If we aren't, should be fine :).

karelhala:
@himdel no params, just fetching JSON data.

himdel:
:+1: Any idea about the failing specs?

karelhala:
@himdel those are related to https://github.com/ManageIQ/ui-components/pull/112

PR #121, opened 2017-08-19
Wrapper component around patternfly-bootstrap-treeview for explorers

skateman:
This is nothing but a wrapper around [patternfly-bootstrap-treeview](https://github.com/patternfly/patternfly-bootstrap-treeview/) to provide it as a component. In the future it should be rewritten without jQuery. It will be used on [left-side explorer screens](https://github.com/ManageIQ/manageiq-ui-classic/issues/1871) in the classic UI. For the more complex trees with checkboxes a new component will be created in the future.

screenshot from 2017-08-19 13-22-20

**Usage:** 
```html
<miq-tree-view name="tree-name" reselect="bool" data="[]" lazy-load="func" on-select="func"></miq-tree-view>
```
The `data` should contain the tree nodes in the format specified by [patternfly-bootstrap-treeview](https://github.com/patternfly/patternfly-bootstrap-treeview/#data-structure), the two callbacks functions should be in the following form:
```js
function lazyLoad(node) {
  return new Promise(function (resolve) {
    // do something  with the node
    resolve(data); // same format as above
  });
}

function nodeSelect(node) {
  // do something with the node
}
```
If you want to allow firing an onSelect event on an already selected node, you should set the `reselect` parameter to `true`.

@karelhala and @himdel please review, @Hyperkid123 you can use this in the dialog editor.

### DIFF
--- a/src/tree-view/treeViewComponent.ts
+++ b/src/tree-view/treeViewComponent.ts
@@ -1,0 +1,79 @@
+import * as ng from 'angular';
+
+export class TreeViewController {
+  public tree;
+  public reselect;
+  public data;
+  public name : string;
+  public onSelect: (args: {node: any}) => void;
+  public lazyLoad: (args: {node: any}) => Promise<any>;
+
+  /*@ngInject*/
+  constructor(private $element : ng.IRootElementService) {}
+
+  public $onInit() {
+    let element = this.$element[0].querySelector('div.treeview');
+    this.renderTree(element).then(() => {
+      this.tree = ng.element(element).treeview(true);
+
+      this.tree.getNodes().forEach((node) => {
+        if (this.getTreeState(node) === !node.state.expanded) {
+          this.tree.toggleNodeExpanded(node);
+        }
+      });
+    });
+  }
+
+  private renderTree(element) {
+    return new Promise((resolve) => {
+      ng.element(element).treeview({
+        data:            this.data,
+        expandIcon:      'fa fa-fw fa-angle-right',
+        collapseIcon:    'fa fa-fw fa-angle-down',
+        loadingIcon:     'fa fa-fw fa-spinner fa-pulse',
+        levels:          1,
+        allowReselect:   this.reselect,
+        showBorders:     false,
+        onNodeExpanded:  this.setTreeState(true),
+        onNodeCollapsed: this.setTreeState(false),
+        onNodeSelected:  (_event, node) => this.onSelect({node: node}),
+        lazyLoad:        (node, render) => this.lazyLoad({node: node}).then(render),
+        onRendered:      () => resolve()
+      });
+    });
+  }
+
+  private setTreeState(state) {
+    return (_event, node) => {
+      let persist = JSON.parse(sessionStorage.getItem(`treeView-${this.name}`));
+      // Initialize the session storage object
+      if (!persist) {
+        persist = {};
+      }
+      // Save the third argument as the new node state
+      persist[node.key] = state;
+      sessionStorage.setItem(`treeView-${this.name}`, JSON.stringify(persist));
+    };
+  }
+
+  private getTreeState(node) {
+    let persist = JSON.parse(sessionStorage.getItem(`treeView-${this.name}`));
+    // Initialize the session storage object
+    if (!persist) {
+      persist = {};
+    }
+    return persist[node.key];
+  }
+}
+
+export default class TreeView implements ng.IComponentOptions {
+  public controller = TreeViewController;
+  public template = '<div class="treeview treeview-pf-select"></div>';
+  public bindings: any = {
+    name: '@',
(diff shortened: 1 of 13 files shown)


src/tree-view/treeViewComponent.ts
```diff
@@ -0,0 +1,81 @@
+const $ = (<any> window).$;
+const Promise = (<any> window).Promise;
```
karelhala: No need to use these constants. `$` is registered via webpack as public package, and Promise is ES6 thing which comes with typescript.

src/tree-view/treeViewComponent.ts
```diff
@@ -0,0 +1,81 @@
...
+export class TreeViewController {
+  public tree;
+  public reselect;
+  public data;
+  public name : string;
+  public onSelect;
+  public lazyLoad;
+
+  /*@ngInject*/
+  constructor(private $element, private $timeout, private $http) {}
```
karelhala: `$timeout` and `$http` is not used in this component aren't they? So we can remove them.

src/tree-view/treeViewComponent.ts
```diff
@@ -0,0 +1,81 @@
...
+        if (this.getTreeState(node) === !node.state.expanded) {
+          this.tree.toggleNodeExpanded(node);
+        }
+      });
+    });
+  }
+
+  private renderTree(element) {
+    return new Promise((resolve) => {
+      $(element).treeview({
```
karelhala: No need to use `$(element)`, angular.element is good enough.
```diff
diff --git a/src/tree-view/treeViewComponent.ts b/src/tree-view/treeViewComponent.ts
index 1cddced..9c36be3 100644
--- a/src/tree-view/treeViewComponent.ts
+++ b/src/tree-view/treeViewComponent.ts
@@ -1,5 +1,6 @@
 const $ = (<any> window).$;
 const Promise = (<any> window).Promise;
+import * as angular from 'angular';
 
 export class TreeViewController {
   public tree;
@@ -27,7 +28,7 @@ export class TreeViewController {
 
   private renderTree(element) {
     return new Promise((resolve) => {
-      $(element).treeview({
+      angular.element(element).treeview({
         data:            this.data,
         expandIcon:      'fa fa-fw fa-angle-right',
         collapseIcon:    'fa fa-fw fa-angle-down',
```

src/tree-view/treeViewComponent.ts
```diff
@@ -0,0 +1,81 @@
...
+        onNodeSelected:  (_event, node) => this.onSelect(node),
+        lazyLoad:        (node, render) => this.lazyLoad(node).then(render),
+        onRendered:      () => resolve()
+      });
+    });
+  }
+
+  private setTreeState(state) {
+    return (_event, node) => {
+      let persist = JSON.parse(sessionStorage.getItem('treeView-' + this.name));
```
karelhala: Instead of using concat use [Template_literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)
```typescript
let persist = JSON.parse(sessionStorage.getItem(`treeView-${this.name}`));
```

src/tree-view/treeViewComponent.ts
```diff
@@ -0,0 +1,81 @@
...
+      persist = {};
+    }
+    return persist[node.key];
+  }
+}
+
+export default class TreeView implements ng.IComponentOptions {
+  public controller = TreeViewController;
+  public template = '<div class="treeview treeview-pf-select"></div>';
+  public controllerAs = '$ctrl';
```
karelhala: `$ctrl` is default value for controllerAs, if we are not going to use it in component, no need to set value to it.

src/tree-view/treeViewComponent.ts
```diff
@@ -0,0 +1,81 @@
...
+        expandIcon:      'fa fa-fw fa-angle-right',
+        collapseIcon:    'fa fa-fw fa-angle-down',
+        loadingIcon:     'fa fa-fw fa-spinner fa-pulse',
+        levels:          1,
+        allowReselect:   this.reselect,
+        showBorders:     false,
+        onNodeExpanded:  this.setTreeState(true),
+        onNodeCollapsed: this.setTreeState(false),
+        onNodeSelected:  (_event, node) => this.onSelect(node),
+        lazyLoad:        (node, render) => this.lazyLoad(node).then(render),
```
karelhala: Even tho it's nice to read promises, can we expose public API with promises? I am okay with local calling of promises, but public API shouldn't expose such thing. Consider changing `lazyLoad` to be [callback_binding](http://www.codelord.net/2016/05/13/understanding-angulars-and-binding/). Same applies to `onSelect` it's not a object, but `callback_binding`.

karelhala:
@skateman looks good, @himdel any objections?

PR #123, opened 2017-08-20
[WIP] Dialog editor entry point treeview

Hyperkid123:
This is a component that enables to choose entry point from a tree view. 
At this point it uses the older jQuery treeview. The goal is to make treeview proper angular component. It should be a dumb component, but since it uses data from tree builder, and lazy loading, it might be practical to keep data loading inside of this component. 

files:
/src/dialog-tree


Hyperkid123:
@miq-bot add_label wip

Hyperkid123:
@miq-bot assign martinpovolny

Hyperkid123:
@miq-bot remove_label wip

demo/services/availableComponentsService.ts
```diff
@@ -93,6 +93,12 @@ export default class AvailableComponentsService {
           'TreeView',
           '/tree-view',
           require('./../views/tree-view/basic.html'), 'demoTreeView as vm')
+      ]),
+      new AvailableGroup('dialog-user', 'Dialog-User Component', '/dialog-user', [
+        new AvailableComponent('dialog-user',
+          'Dialog-user',
+          '/dialog-user',
+          require('./../views/dialog/user.html'), 'demoDialogs as vm')
```
skateman: You don't need this, it's already there on lines 86-89

skateman:
@himdel @karelhala some new files appeared in the `dist` folder, is that OK here?

demo/services/availableComponentsService.ts
Hyperkid123: deleted :+1:

karelhala:
@skateman that's not ok. Those are fonts which should not be pushed to git. However there is missing rule in `.gitignore`

skateman:
@Hyperkid123 let's extract the `dialog-tree` into a more generic `tree-selector` component with a `selectables` regexp attribute that can specify what can be selected.

Hyperkid123:
@miq-bot add_label wip

himdel:
gitignoring fonts solved in https://github.com/ManageIQ/ui-components/pull/126 :)

PR #126, opened 2017-08-24
Added $timeout to tree-view functions

Hyperkid123:
Added $timeout to treeview, so wrapper components don't have to lazy-load, and node-selected functions.


Hyperkid123:
@miq-bot assign himdel

PR #127, opened 2017-08-24
Add available component builder so we can use it much easilly

karelhala:
To register new group you will have to add new item with name, location, title and empty array of components. When registering available component use custom decorator above some class to register it:
```Typescript
@ComponentDemo({
  name: 'tree-view-new',
  title: 'TreView',
  template: require('./../views/tree-view/basic.html'),
  controller: 'demoTreeView as vm'
})
export default class TreeViewController {
  //Demo treeView controller implementation
}
```

If you need more components registered under same controller use `AvailableCmp` multiple times as seen at [demo/controllers/dataTableController.ts](https://github.com/ManageIQ/ui-components/pull/127/files#diff-5884d947210a1d2824f3a7a10e833088)
```Typescript
@ComponentDemo({
  name: 'small',
  title: 'Small tile',
  template: require('./../views/tile-view/small.html'),
  group: 'tile-menu',
  controller: 'demoDataTable as vm'
})
@ComponentDemo({
  name: 'big',
  title: 'Big tile',
  template: require('./../views/tile-view/big.html'),
  group: 'tile-menu',
  controller: 'demoDataTable as vm'
})
@ComponentDemo({
  name: 'basic',
  title: 'Basic data table',
  template: require('./../views/data-table/basic.html'),
  group: 'data-table',
  controller: 'demoDataTable as vm'
})
export default class DataTableController {
  //Implementation of DataTablecontroller
}
```

### DIFF
--- a/demo/services/availableComponentBuilder.ts
+++ b/demo/services/availableComponentBuilder.ts
@@ -1,0 +1,32 @@
+import {AvailableComponent, default as AvailableComponentsService} from './availableComponentsService';
+
+function checkProperties(settings, fields, target?) {
+  for (let field of fields) {
+    if (!settings.hasOwnProperty(field)) {
+      let suffix = target ? `Called on class ${target['name']}.` : '';
+      throw new Error(`Available component has to have ${field} set up! ${suffix}`);
+    }
+  }
+}
+
+export function ComponentDemo(settings: any) {
+  function componentFactory() {
+    return new AvailableComponent(
+      settings.name,
+      settings.title || settings.name,
+      settings.location || `/${settings.name}`,
+      settings.template,
+      settings.controller
+    );
+  }
+
+  return function(target: Function) {
+    checkProperties(settings, ['controller', 'name', 'template', 'group'], target);
+    let availCmps = new AvailableComponentsService();
+    const currentGroup = availCmps.availableComponents.filter(group => group.name === settings.group);
+    if (!currentGroup) {
+      throw new Error(`Selected group '${settings.group}' does not exists!`);
+    }
+    currentGroup.forEach(oneGroup => oneGroup.components.push(componentFactory()));
+  };
+}
--- a/demo/controllers/availableComponentsController.ts
+++ b/demo/controllers/availableComponentsController.ts
@@ -3,6 +3,7 @@ import {IAvailableGroup} from '../services/availableComponentsService';
 
 export default class AvailableComponentsController {
   public availableComponents: IAvailableGroup[];
+
   /* @ngInject */
   public constructor() {
     this.availableComponents = (new AvailableComponentsService()).availableComponents;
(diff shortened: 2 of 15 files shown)


karelhala:
@skateman @mtho11 @himdel @Hyperkid123 what you think?

skateman:
Well, for me it's just more unreadable :disappointed:

karelhala:
It's just a suggestion, if you do not like we don't have to use. But @Hyperkid123 had problem with creating new component and he had some item missing so I thought someone might like this approach.

Hyperkid123:
This seems to be much better solution for people with goofy eyes (like me).

skateman:
Can't we ruby-ize this by dynamically loading all the html files that will have some kind of special comments at the beginning of each file? Like this:
/*
 * controller: XY controller as vm
 * url: bleh
 * section: XY components
 * name: XY1
 */

himdel:
I like the idea, but hate that syntax... (and having to write a method for each param)

What about something simpler like..

```js
groupFactory({
  name: 'tree-view',
  title: 'Tree Components',
  location: '/tree',
  components: [
    componentFactory({
      controller: 'demoTreeView as vm',
      location: '/tree-view-new',
      name: 'tree-view-new',
      title: 'TreView',
      template: require('./../views/tree-view/basic.html'),
    }),
  ],
});
```

karelhala:
Truth to be told the it's not easy to read either. What would you say about Ts decorators? Above example would be written like:
```Typescript
@AvailableCmp({
  name: 'tree-view-new',
  title: 'TreView',
  template: require('./../views/tree-view/basic.html'),
  controller: 'demoTreeView as vm'
})
export default class TreeViewController {
  //Demo treeView controller implementation
}
```

karelhala:
@himdel @Hyperkid123 @mtho11 what do you say about these custom decorators?

himdel:
Just one question .. what is AvailableCmp? :)

Wouldn't something like `ComponentDemo` make more sense?

(I mean, I get that the menu is called available components but .. that name doesn't seem to have anything to do with what this does, does it?)

karelhala:
> Wouldn't something like ComponentDemo make more sense?

Good point. Renaming it to `ComponentDemo` decorator.

Hyperkid123:
I like it

demo/services/availableComponentBuilder.ts
```diff
@@ -0,0 +1,32 @@
+import {AvailableComponent, default as AvailableComponentsService} from './availableComponentsService';
```
skateman: You probably want to rename this accordingly to the new name :wink:

karelhala: That's ok, it's used in `componentFactory` and in decorator function.

karelhala:
@himdel @skateman what do you think about it?

himdel:
@karelhala LGTM, is this ready to merge? :)

But .. can you please update the description to fit the new names?

karelhala:
@himdel updated

himdel:
Merging, this should only affect the demo - no release concerns, right?

PR #129, opened 2017-08-24
Add one-way binding for external tree node selection in treeView

skateman:
If you want to programmatically select a node from outside the component, simply set its `selected` attribute to an object that has some matching attributes with a node.

`<miq-tree-view ... selected="{key: 'something'}">`

@karelhala or @himdel


skateman:
@karelhala I think this is ready to review & merge

karelhala:
LGTM

PR #130, opened 2017-08-24
[WIP] Make sure parent nodes are expanded when loading treeView state

skateman:
There's no way to reproduce the issue using clicking, but #129 introduces a programmatic way of selecting and expanding tree nodes. If a non-visible node with children gets selected and expanded, the tree state object will hold its state and after a refresh it can go ugly (see below).

This PR enforces to expand all the parent nodes of an expanded node.

**Before:**
screenshot from 2017-08-24 17-19-43

**After:**
screenshot from 2017-08-24 17-20-02


skateman:
Closing this and pushing it to #129

PR #131, opened 2017-08-26
Allow to select a lazily loadable node in treeView

skateman:
Sometimes we want to select a node that's not yet loaded. This is now possible by passing an array of node matchers in the `selected` attribute of the component. The first element of the array should match a node that's already available in the tree. The others should match its children, each in one level until the last element that matches the finaly node to be selected.

For example:
```json
[
  {
    "title": "Parent",
    "nodes": [
      {
        "title": "Child",
        "nodes": [
          {
            "title": "Grandchild"
          }
        ]
      }
    ]
  }
]
```
If the tree above is lazily loaded on each level and you want to select the `Grandchild` node, you would need:
```html
<miq-tree-view ... selected="[{'title': 'Parent'}, {'title': 'Child'}, {'title': 'Grandchild'}]"/>
```

### DIFF
--- a/demo/controllers/treeViewController.ts
+++ b/demo/controllers/treeViewController.ts
@@ -6,13 +6,15 @@ export default class TreeViewController {
   public selectNode;
 
   /*@ngInject*/
-  constructor(private $scope : ng.IScope, private $timeout : ng.ITimeoutService) {};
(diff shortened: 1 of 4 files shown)


karelhala:
LGTM, merging.

PR #133, opened 2017-08-29
Improved state (re)storing for the TreeView component

skateman:
I extracted the miq-specific `key` attribute from the state storing part of TreeView. From now on if you want to store the tree state, you'd need to set the `persist` string parameter to a key that's uniquely set for each node in the `data`. In the miq-related context this should be `key`, but it basically can be anything. It was not necessary to store collapsed nodes in the session storage as they are always rendered as collapsed. So now it's possible to save the path to an expanded node instead of the `true` value that helps us in restoring the state of a lazily loadable node.

### DIFF
--- a/demo/views/tree-view/basic.html
+++ b/demo/views/tree-view/basic.html
@@ -3,7 +3,14 @@
 <button ng-click="vm.selectLazy();">Select lazy</button>
 <p>Note: programmatically selecting a node is not firing the onSelect event on purpose.</p>
 
-<miq-tree-view name="demo-tree" data="vm.data" lazy-load="vm.lazyLoad(node)" on-select="vm.nodeSelect(node)" selected="vm.selectNode"></miq-tree-view>
+<miq-tree-view
+  name="demo-tree"
+  data="vm.data"
+  lazy-load="vm.lazyLoad(node)"
+  on-select="vm.nodeSelect(node)"
+  selected="vm.selectNode"
+  persist="key"
+></miq-tree-view>
 
 <div ng-if="vm.node">
   <h3>Selected node:</h3>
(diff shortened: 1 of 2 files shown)


demo/views/tree-view/basic.html
```diff
@@ -3,7 +3,7 @@
 <button ng-click="vm.selectLazy();">Select lazy</button>
 <p>Note: programmatically selecting a node is not firing the onSelect event on purpose.</p>
 
-<miq-tree-view name="demo-tree" data="vm.data" lazy-load="vm.lazyLoad(node)" on-select="vm.nodeSelect(node)" selected="vm.selectNode"></miq-tree-view>
+<miq-tree-view name="demo-tree" data="vm.data" lazy-load="vm.lazyLoad(node)" on-select="vm.nodeSelect(node)" selected="vm.selectNode" persist="key"></miq-tree-view>
```
karelhala: Can you please format it a bit better?
```html
<miq-tree-view name="demo-tree"
                         data="vm.data"
                         lazy-load="vm.lazyLoad(node)"
                         on-select="vm.nodeSelect(node)"
                         selected="vm.selectNode"
                         persist="key"></miq-tree-view>
```

src/tree-view/treeViewComponent.ts
```diff
@@ -83,73 +82,132 @@ export class TreeViewController {
...
-    let persist = JSON.parse(sessionStorage.getItem(`treeView-${this.name}`));
-    // Initialize the session storage object
-    if (!persist) {
-      persist = {};
-    }
-    return persist[node.key];
+  private loadTreeState() {
+    let store = JSON.parse(sessionStorage.getItem(`treeView-${this.name}`)) || {};
+    // Create a list of store keys that should be ignored
+    let blacklist = _.flatten(Object.keys(store).map(key => store[key])).map(obj => obj[this.persist]);
```
karelhala:
```typescript
let blacklist = _.flatten(Object.keys(store)
                  .map(key => store[key]))
                  .map(obj => obj[this.persist]);
```

karelhala:
LGTM

PR #137, opened 2017-08-31
Dialog Editor Validations

romanblanco:
taking the method from PR https://github.com/ManageIQ/manageiq-ui-classic/pull/2055 as a new service for Dialog validations:

> 
> After these changes, it won't be possible to submit invalid dialog.
> The methods are testing that
> 
> - dialog has a name
> - dialog has at least one tab
> - every dialog tab have a name
> - every dialog tab has at least one box
> - every dialog box has a name
> - every dialog box has at least one element
> - every element has a name
> 
> if one test fails, the `Save` buttons becomes disabled.
> 
> Links
> ----------------
> * https://www.pivotaltracker.com/story/show/144481043
> 
> Steps for Testing/QA
> -------------------------------
> 
> Automation -> Automate -> Customization -> Edit this / Add a new dialog in the Dialog Editor -> [see the screencast]
> 
> screencast from 2017-08-30 16-21-22

/cc @skateman @himdel @karelhala

### DIFF
--- a/src/dialog-editor/services/dialogValidationService.ts
+++ b/src/dialog-editor/services/dialogValidationService.ts
@@ -1,0 +1,65 @@
+import {__} from '../../common/translateFunction';
+import * as _ from 'lodash';
+
+export default class DialogValidationService {
+  public invalid: any = [];
+
+  /** @ngInject */
+  constructor(private DialogEditor: any) {
+  }
+
+  /**
+   * Run validations across each dialog elements.
+   * @memberof DialogValidationService
+   * @function dialogIsValid
+   */
+  public dialogIsValid() {
+    var validators = {
+      dialog: [
+        dialog => ({ status: ! _.isEmpty(dialog.label),
+                     errorMessage: __('Dialog needs to have a label') }),
+        dialog => ({ status: dialog.dialog_tabs.length > 0,
+                     errorMessage: __('Dialog needs to have at least one tab') })
+      ],
+      tabs: [
+        tab => ({ status: ! _.isEmpty(tab.label),
+                  errorMessage: __('Dialog tab needs to have a name') }),
+        tab => ({ status: tab.dialog_groups.length > 0,
+                  errorMessage: __('Dialog tab needs to have at least one box') })
+      ],
+      groups: [
+        group => ({ status: ! _.isEmpty(group.label),
+                    errorMessage: __('Dialog box needs to have a name') }),
+        group => ({ status: group.dialog_fields.length > 0,
+                    errorMessage: __('Dialog box needs to have at least one element') })
+      ],
+      fields: [
+        field => ({ status: ! _.isEmpty(field.name),
+                    errorMessage: __('Dialog element needs to have a name') }),
+        field => ({ status: ! _.isEmpty(field.label),
+                    errorMessage: __('Dialog element needs to have a label') })
+      ]
+    };
+
+    let invalid = [];
+    let validate = function (f, item) {
+      let validation = f(item);
+      if (!validation.status) {
+        invalid.push({ message: validation.errorMessage, item: item });
+      }
+      return validation.status;
+    };
+
+    return _.every(DialogEditor.data.content, dialog =>
+      _.every(validators.dialog, f => validate(f, dialog)) &&
+      _.every(dialog.dialog_tabs, tab =>
+        _.every(validators.tabs, f => validate(f, tab)) &&
+        _.every(tab.dialog_groups, group =>
+          _.every(validators.groups, f => validate(f, group)) &&
+          _.every(group.dialog_fields, field =>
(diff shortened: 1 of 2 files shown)


src/dialog-editor/services/dialogValidationService.ts
```diff
@@ -0,0 +1,66 @@
+import {__} from '../../common/translateFunction';
+import * as _ from 'lodash';
+
+export default class DialogValidationService {
+  public invalid: any = {};
+
+  /** @ngInject */
+  constructor(private DialogEditor: any) {
```
skateman: Is this a good idea? Is this even possible?

romanblanco: @skateman I need the data from the `DialogEditor` service. Tested and works fine. Any idea would you like to improve it?

skateman: Is this REALLY a service if it provides data? I'm not THAT familiar with angular but AFAIK services are singletons that should be used/called in controllers and the data should be passed as an argument and not as a dependency. But maybe I'm wrong so pinging @himdel @karelhala

himdel: Probably too late for such a design change now. But.. using services as a data/state store is quite common in angular world.

(I guess it's not yet the time to move this to redux, right?)

himdel: Ah.. but it's not too late not to rely on DialogEditor here, we can just pass the data as an argument to `dialogIsValid` - cleaner that way probably :)

src/dialog-editor/services/dialogValidationService.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * Run validations across each dialog elements.
+   * @memberof DialogValidationService
+   * @function dialogIsValid
+   */
+  public dialogIsValid() {
```
himdel: `public dialogIsValid(content: any) {` ?

src/dialog-editor/services/dialogValidationService.ts
```diff
@@ -0,0 +1,66 @@
...
+    let validators = {
```
himdel: Hm, should this be hardcoded here?

I'm not coming up with any good examples of wanting to add custom validations to the dialog *editor* so maybe the answer is yes :). But is it? :)

src/dialog-editor/services/dialogValidationService.ts
karelhala: Forgot to add my comment before. To store current data in service is OK to do, however while we are thinking to use redux, we should reimplement the `dialogIsValid` function to accept dialog data. That way once we move this to use redux we can control them by redux and they will be passed to this function every time they change.

src/dialog-editor/services/dialogValidationService.ts
karelhala: It would be nice to add new validators, so @romanblanco care to rewrite this like this?
```Typescript
export default class DialogValidationService {
  private validators: any;
  constructor(private DialogEditor: any) {
    this.validators = {
      dialog: DialogValidationService.initDialogValidators(),
      tabs: DialogValidationService.initTabsValidators(),
      groups: DialogValidationService.initGroupsValidators()
      fields: DialogValidationService.initFieldsValidators()
    }
  }
  private static initGroupsValidators() {
    return [
      group => ({ status: ! _.isEmpty(group.label),
                    errorMessage: __('Dialog box needs to have a name') }),
      group => ({ status: group.dialog_fields.length > 0,
                    errorMessage: __('Dialog box needs to have at least one element') })
    ]
  }
  private static initFieldsValidators() {
    return [
      field => ({ status: ! _.isEmpty(field.name),
                    errorMessage: __('Dialog element needs to have a name') }),
      field => ({ status: ! _.isEmpty(field.label),
                    errorMessage: __('Dialog element needs to have a label') })
    ]
  }
  private static initTabsValidators() {
    return [
      tab => ({ status: ! _.isEmpty(tab.label),
                  errorMessage: __('Dialog tab needs to have a name') }),
      tab => ({ status: tab.dialog_groups.length > 0,
                  errorMessage: __('Dialog tab needs to have at least one box') })

    ]
  }
  private static initDialogValidators() {
    return [
      dialog => ({ status: ! _.isEmpty(dialog.label),
                     errorMessage: __('Dialog needs to have a label') }),
      dialog => ({ status: dialog.dialog_tabs.length > 0,
                     errorMessage: __('Dialog needs to have at least one tab') })
    ]
  }
}
```

romanblanco:
@skateman @karelhala @himdel updated the PR as we discussed

romanblanco:
BZ ticket fixed by this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1487774

PR #138, opened 2017-08-31
Add ability to edit the dialog field associations in the UI

eclarizio:
Note that users will no longer be able to turn on "auto refresh" or "auto refresh other fields when modified" after this PR, as the dialog field associations are going to end up handling that logic. The classic UI dialog editor will still retain these options for now.

This will require [this PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/2112) from the classic UI to ignore the list of dynamic dialog fields that is going to be passed through the API, as well as [this PR](https://github.com/ManageIQ/manageiq/pull/15937) from the main repo to actually build the associations from the IDs being passed in.

https://www.pivotaltracker.com/story/show/148013513

/cc @gmcculloug, @romanblanco 

### DIFF
--- a/src/dialog-editor/components/modal-field-template/check-box.html
+++ b/src/dialog-editor/components/modal-field-template/check-box.html
@@ -21,10 +21,3 @@
          switch-on-text="{{'Yes'|translate}}"
          switch-off-text="{{'No'|translate}}">
 </div>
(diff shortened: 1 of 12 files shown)


martinpovolny:
@romanblanco : can you review this one, please?

PR #142, opened 2017-09-01
Do not render the TreeView before it receives the node data

skateman:
Do not render the tree until its `data` attribute is set. This will allow us to lazily load the tree after the component has been displayed.


karelhala:
LGTM

PR #145, opened 2017-09-05
Allow reloading the TreeView if its data has been changed

skateman:
I thought we can avoid having this feature, but it seems like we're re-rendering the trees sometimes.


martinpovolny:
ping @karelhala

karelhala:
I hope the removing of tree and re rendering is just temporary solution. Otherwise looks good.

skateman:
@karelhala it's a necessary evil until we do some refactoring in the explorer area. But until then the feature is needed for a bugfix...

PR #147, opened 2017-09-05
treeview - mark optional bindings optional

himdel:
`selected` & `reselect` were already optional
`onSelect` & `lazyLoad` are now :)

Cc @skateman

### DIFF
--- a/src/tree-view/treeViewComponent.ts
+++ b/src/tree-view/treeViewComponent.ts
@@ -70,8 +70,8 @@ export class TreeViewController {
         showBorders:     false,
         onNodeExpanded:  this.storeNodeState(true),
         onNodeCollapsed: this.storeNodeState(undefined),
-        onNodeSelected:  (_event, node) => this.$timeout(() => this.onSelect({node: node})),
-        lazyLoad:        (node, render) => this.$timeout(() => this.lazyLoad({node: node}).then(render)),
+        onNodeSelected:  this.onSelect ? (_event, node) => this.$timeout(() => this.onSelect({node: node})) : () => null,
+        lazyLoad:        this.lazyLoad ? (node, render) => this.$timeout(() => this.lazyLoad({node: node}).then(render)) : () => null,
         onRendered:      () => this.$timeout(resolve)
       });
     }).then(() => {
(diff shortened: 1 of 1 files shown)


src/tree-view/treeViewComponent.ts
```diff
@@ -44,8 +44,8 @@ export class TreeViewController {
         showBorders:     false,
         onNodeExpanded:  this.storeNodeState(true),
         onNodeCollapsed: this.storeNodeState(undefined),
-        onNodeSelected:  (_event, node) => this.$timeout(() => this.onSelect({node: node})),
-        lazyLoad:        (node, render) => this.$timeout(() => this.lazyLoad({node: node}).then(render)),
+        onNodeSelected:  this.onSelect ? (_event, node) => this.$timeout(() => this.onSelect({node: node})) : () => null,
```
skateman: These lines are too long for the linter :cry:

skateman: @karelhala can't those be set as default values at the type declaration?

karelhala: Either better space align or, we can use private methods for these 2 functions
```typescript
  private onSelected(node) {
    this.$timeout(() => this.onSelect({node: node}));
  }

  private onLazyLoad(node, render) {
    this.$timeout(() => this.lazyLoad({node: node}).then(render));
  }

  private renderTree() {
    new Promise((resolve) => {
      this.element.treeview({
        //...
        onNodeSelected:  this.onSelect ? (_event, node) => this.onSelected(node) : () => null,
        lazyLoad:        this.lazyLoad ? (node, render) => this.onLazyLoad(node, render) : () => null,
        //...
      }
    }
  }
```

@himdel what do you think?

himdel: Well, I think `$timeout` needs to live in that `treeview` call and not somewhere else.. and after that, there's not much point in creating functions.

So.. I'd just disable the long lines linter rule :).

But.. what aobut just providing a default value for `onSelect` and `lazyLoad`? That way, no conditionals necesarry...

skateman: Default values :+1:, you have to put them into the constructor as it doesn't work at the declaration (tried, failed)

karelhala:
Looks good, please rebase to master.

himdel:
Rebased, thanks :)

PR #149, opened 2017-09-06
tileView - fix columns to actually be available inside the card

himdel:
the card could not access `tileCtrl.columns` because it has access only to the `pf-card-view` `config` object.

Adding `columns` to the config object so that it is available.

Also fixed the spacing a bit - there was no visual space between `dt` and `dl` (**Foo**:bar)

Cc @karelhala 

Infra provider - fine & now:

provider-infra

fix-provider-infra

https://bugzilla.redhat.com/show_bug.cgi?id=1486702


karelhala:
Both JS and CSS changes looks good. Merging.

PR #151, opened 2017-09-07
Changed background color in the Dialog Editor

romanblanco:
The gray background looked strange in the editor, trying white colour

Before:
screenshot-2017-9-7 manageiq automation

After:
screenshot-2017-9-7 manageiq automation 1

@dtaylor113 what you think?

### DIFF
--- a/src/styles/dialog-editor.scss
+++ b/src/styles/dialog-editor.scss
@@ -1,5 +1,5 @@
 .dialog-designer-container {
-  background-color: #f1f1f1;
+  background-color: #ffffff;
   height: 100%;
   width: 100%;
 


dtaylor113:
@romanblanco, looks fine.  Only little issue is that the "(+) Create Tab" gets a little washed out and doesn't stand out as an operation (as opposed to a Tab).   But I like this better than the grey bkground.  @serenamarie125

romanblanco:
@dtaylor113 With gray colour the tabs were getting 'washed out', plus with showing the 'Close' and 'Edit' buttons on hover it was a bit difficult to change a tab and not closing it 😄 

screencast from 2017-09-07 15-42-06

It still is a bit difficult when the dialog has more tabs, but with the white colour it's a bit better, but still I'm not sure that showing 'Close' and 'Edit' buttons at tab after hover is the best solution.

dtaylor113:
Hmm...so the active tab is the only one which is white.  Seems like there should be a difference between non-active tabs and the background.

romanblanco:
@martinpovolny can you merge this?

himdel:
@romanblanco Any plans on addressing https://github.com/ManageIQ/ui-components/pull/151#issuecomment-327834862 ?

romanblanco:
@himdel Not sure how, so probably not in this PR
/cc @dtaylor113

himdel:
Understood, merging :)

PR #155, opened 2017-09-12
Use font-fabulous from the NPM repo instead of GitHub

skateman:
@himdel this `yarn.lock` thing is scaring me, can you please review?


himdel:
Yeah, this kind of diff garbage is exactly why we have it in `.gitignore` in ui-classic (for non-release branches).

I wonder if it makes sense to do the same thing here, and just generate it for npm releases.. @karelhala ?

karelhala:
@himdel I am all for it, to be removed. I don't know if it's critical or not, but we can use npm's lock file. But again, I don't think lock files are needed for JS stuff.

PR #159, opened 2017-09-18
Fix issue in dialog-user where values for dates were undefined

eclarizio:
In the classic UI, there was an issue when using a date or date/time field where if there is not a default value set, it was passing in an undefined attribute for `.values`, and thus the ui-component was crashing trying to do `new Date(data.values)`.

This should fix that issue and just initialize the date or date/time field to the current date and time.

@chalettu Can you review?


eclarizio:
@miq-bot assign @chriskacerguis 

Hey Chris, just assigning this to you for now, if you want to re-assign please do. Otherwise I'm not sure why the coverage remained the same, since I added more tests.

PR #163, opened 2017-09-21
Remove CSS rule to hide help icon in dialog fields

romanblanco:
screencast from 2017-09-21 10-33-13

**Pivotal tasks**:

https://www.pivotaltracker.com/story/show/150614285
https://www.pivotaltracker.com/story/show/150614221

@chalettu can you take a look on this?

### DIFF
--- a/demo/data/dialog-data.json
+++ b/demo/data/dialog-data.json
@@ -100,6 +100,7 @@
                       "created_at": "2017-06-16T19:29:28Z",
                       "updated_at": "2017-06-16T19:29:28Z",
                       "label": "Environment",
+                      "description": "This shows the help message for the field",
                       "dialog_group_id": 10000000002378,
                       "position": 1,
                       "dynamic": false,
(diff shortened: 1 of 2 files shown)


Loicavenel:
@serenamarie125 can you review this one?

chalettu:
Actually.  This was hidden on purpose because we are waiting on an API addition for field level help .  The description field is not the same as the new field that they will be adding.

romanblanco:
@chalettu I've understood, that the _description_ should be just replaced by _help_, because the _description_ is not actually much used. Or do we want to have both _description_ and _help_ at the same time? @Loicavenel

dclarizio:
@chalettu it's my understanding that @Loicavenel decided to re-use the `description` field to hold the online help text.  There was even a migration done to expand the length of text it can hold.
Will let Loic chime in to verify.

himdel:
Looks like this just unhides the help, and updates the field which is currently being displayed.

Another PR depends on this, merging, we can always rename the field later.

PR #164, opened 2017-09-25
Split up isIconOrImage helper function into hasIcon and hasImage

skateman:
In the future we want to favor images over icons and testing which item has which one is easier this way.


karelhala:
Looks good! Merging.

PR #168, opened 2017-09-26
Align the page indication number input field to center

skateman:
After consulting with @sbulage we agreed that the fix for this BZ should be to align the content of the input field to center. However, I'm not sure if this is good from the UX perspective, can you please take a look @epwinchell?

**Before:**
screenshot from 2017-09-26 10-59-18

**After:**
screenshot from 2017-09-26 10-59-06

https://bugzilla.redhat.com/show_bug.cgi?id=1486644

### DIFF
--- a/src/styles/ui-components.scss
+++ b/src/styles/ui-components.scss
@@ -360,7 +360,7 @@ miq-fonticon-picker {
       margin: 0;
     }
     input[type="number"] {
-      text-align: right;
+      text-align: center;
       width: 35px;
       -moz-appearance: textfield;
     }


epwinchell:
@skateman According to PF, it was aligned correctly (right-justified.)

http://www.patternfly.org/pattern-library/navigation/pagination/)

martinpovolny:
@serenamarie125, @epwinchell : shall we close this one?

skateman:
@martinpovolny I'd totally close it together with the BZ but let's wait for @serenamarie125

serenamarie125:
According to PatternFly, it should be right aligned.  

@skateman It would be nice to have a 5px margin to the right of the number.  Unsure if we can have the background of that text field be white?  That would also help be more aligned with PF
pagination_callouts

PR #169, opened 2017-09-26
Fixed GTL icons and images when both are present

skateman:
The `hasOwnProperty` wasn't testing if the property is `null` so the `hasImage` and `hasIcon` functions always returned `true`.


karelhala:
Looks good to me. Merging.

PR #170, opened 2017-09-26
Update element fields to match the old Dialog Editor

romanblanco:
~**should be merged after https://github.com/ManageIQ/ui-components/pull/138**~

Fields for the dialog elements have changed in https://github.com/ManageIQ/manageiq-ui-classic/pull/1199

This PR is updating the fields in the new Dialog Editor.

Pivotal Tracker task: https://www.pivotaltracker.com/n/projects/1613907/stories/147779457


romanblanco:
I've created a schema how the fields are changing for myself, maybe it could help with review:



_(fields marked by pink colour are removed in PR #138)_

src/dialog-editor/components/modal-field-template/check-box.html
```diff
@@ -1,30 +1,93 @@
...
+             switch-off-text="{{'No'|translate}}">
+    </div>
+    <div pf-form-group pf-label="{{'Visible'|translate}}">
+      <input bs-switch
+             ng-model="vm.modalData.visible"
+             type="checkbox"
+             switch-on-text="{{'Yes'|translate}}"
+             switch-off-text="{{'No'|translate}}">
+    </div>
+    <div pf-form-group pf-label="{{'Default value'|translate}}">
```
romanblanco: The "Default value" is probably wrong here. @eclarizio

romanblanco:
@miq-bot remove_label unmergeable

romanblanco:
@karelhala please review

himdel:
UI difference:

tbo
tbo2

I guess that's intended :).

Merging.

eclarizio:
@romanblanco @himdel Wait, what, why did the "Fields to refresh" drop down disappear in the after? Did whatever you were testing not contain any dynamic fields?

After looking further, just the template was removed, but in favor of copy/pasting the `dialog-field-association` html everywhere else instead. This should definitely be changed back to use the template.

romanblanco:
@eclarizio I'm not sure what is wrong.

Elements used to have the same fields when they were dynamic, but that has changed.

So I've updated the templates so now whole template for an element is in one file, and as you say, removed the template and copy&pasted the _Fields to refresh_ to every template. So it should be OK. 

Let me know, why to change it back, thanks.

eclarizio:
@romanblanco My point was why delete that file and copy/paste when you could have just included that template in every file? Now if we need to make a change to something in that small code block, it needs to be changed in a bunch of places instead of one.

eclarizio:
@romanblanco Also, it appears you've removed the association drop down from static elements when it should be available for them. There's no reason that a static element shouldn't be able to trigger a refresh for a dynamic field.

romanblanco:
@eclarizio Right. My mistake. I'll fix and let you know. Thanks

PR #172, opened 2017-09-26
Add tree persistence support for TreeSelectorComponent

skateman:
Forgot this feature earlier :disappointed:

### DIFF
--- a/src/tree-selector/treeSelector.html
+++ b/src/tree-selector/treeSelector.html
@@ -1,6 +1,7 @@
 <miq-tree-view
   name="{{ $ctrl.name }}"
   data="$ctrl.parsedData"
+  persist="{{ $ctrl.persist }}"
   selected="$ctrl.selected"
   reselect="true"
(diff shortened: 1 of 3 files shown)


martinpovolny:
So what is this supposed to do? And where are the tests?

skateman:
@martinpovolny it passes the `persist` argument to the underlying [`treeView`](https://github.com/ManageIQ/ui-components/blob/master/src/tree-view/treeViewComponent.ts#L76) component. For now we don't need it but wanted to keep the two components symmetric as possible. Not sure how and why should I test a trivial angular feature (i.e. passing bindings) that's being provided by the framework. I'm working on some tests for both components, but other things first ...

martinpovolny:
@skateman : ok, I saw nothing but the passing of the argument and I was unsure because of "brief" description.

skateman:
@martinpovolny sorry, I thought it's trivial as it is a symmetric feature...

PR #181, opened 2017-10-12
Don't render the treeview before the angular template is fully ready

skateman:
The treeview requires an element ID for the proper styling setup. Without it the styling of nodes might be ignored if you have more than one treeview on a single page. I added a dynamic ID to the component's template, however, it was necessary to delay the tree rendering until the ID is set properly.

The best way to test this is through the *Automate -> Customization -> Buttons* tree, where you can have colored icons.

**Before:**
screenshot from 2017-10-12 13-27-14

**After:**
screenshot from 2017-10-12 13-26-43

https://bugzilla.redhat.com/show_bug.cgi?id=1498052

### DIFF
--- a/src/tree-view/treeViewComponent.ts
+++ b/src/tree-view/treeViewComponent.ts
@@ -20,7 +20,18 @@ export class TreeViewController {
     this.errorHandlers = this.errorHandlers || {};
   }
 
+  public $postLink() {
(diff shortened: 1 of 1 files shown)


skateman:
@miq-bot add_label bug

karelhala:
LGTM, merging

PR #183, opened 2017-10-12
Add es6-shim dependency for development and testing

skateman:
This will allow us to use ES6 features in testing (PhantomJS) and in development (IE11).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "npm": ">= 3.8.1"
   },
   "devDependencies": {
+    "@manageiq/font-fabulous": "^1.0.0",
     "@types/angular-mocks": "^1.5.7",
     "@types/angular-ui-router": "^1.1.35",
     "@types/jasmine": "^2.5.38",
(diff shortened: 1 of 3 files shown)


skateman:
ping @himdel & @karelhala

mtho11:
I must be missing something here as the tests are in typescript which is essentially ES6+. Are there certain features missing in typescript that need this?

skateman:
As @himdel explained to me, we don't have es5 as the compilation target and we're using the same shim in the ui classic.

himdel:
@mtho11 Indeed, both ui-service and ui-classic are using es6-shim, we added `lib: ['es2015', ...]` to `tsconfig.json` to get rid of warnings about code like `[].includes(x)` or `Object.keys(x)`, but it means the demo & specs now work only in es6(lib)-capable browsers.

This fixes that :).

So, essentially `ui-components` as a lib still targets only sane browsers, and it's still up to the project using it to provide such environment, but `ui-components` as a demo will work in older browsers again (and tests won't break in phantomjs).

LGTM

Issue #180, opened 2017-10-11
Dialog field regex not validating


chalettu:
When testing with the regex string  \A[\w+\-.]+@[a-z\d\-]+(\.[a-z]+)*\.[a-z]+\z
the API returns ```\\A[\\w+\\-.]+@[a-z\\d\\-]+(\\.[a-z]+)*\\.[a-z]+\\z```  I think the escaped ```\\``` is causing our regex validation to not work.  Investigate issue and fix.

h-kataria:
closing, already implemented
![Screenshot from 2020-06-30 11-38-36](https://user-images.githubusercontent.com/3450808/86146502-4e68bf80-bac6-11ea-9c7d-1447bd07daea.png)

![Screenshot from 2020-06-30 11-38-19](https://user-images.githubusercontent.com/3450808/86146510-51fc4680-bac6-11ea-9f9a-2e6f219cbf22.png)

Issue #179, opened 2017-10-11
Dialog field regex validation


chalettu:
Per @loicavenel, if a field has a regular expression check defined on it, it should run field validations whether or not the field itself is "required".  The implementation needs to account for allowing blank to be considered as valid input since the field is optional.  It needs to disable submission if someone edits the field and the field doesn't match validation regex.

h-kataria:
already implemented

![Screenshot from 2020-06-30 11-38-36](https://user-images.githubusercontent.com/3450808/86146502-4e68bf80-bac6-11ea-9c7d-1447bd07daea.png)

![Screenshot from 2020-06-30 11-38-19](https://user-images.githubusercontent.com/3450808/86146510-51fc4680-bac6-11ea-9f9a-2e6f219cbf22.png)

Issue #178, opened 2017-10-11
Dialogs and required checkboxes


chalettu:
Per @Loicavenel we expect when a checkbox is "required", that means that it needs to validate as being checked. Any other state should produce an error.

h-kataria:
Closing, already implemented

![Screenshot from 2020-06-30 11-33-32](https://user-images.githubusercontent.com/3450808/86145949-93402680-bac5-11ea-929a-d665f89c830d.png)

PR #184, opened 2017-10-12
Fixed bugs with dialog field validations

chalettu:
@miq-bot add_label bug

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,3 @@
 # dialog editor
 src/dialog-editor @romanblanco
+src/dialog-user @chalettu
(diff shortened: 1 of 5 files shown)


chalettu:
This PR addresses https://github.com/ManageIQ/ui-components/issues/178
https://github.com/ManageIQ/ui-components/issues/179
and https://github.com/ManageIQ/ui-components/issues/180 .  These issues were all inter related which is why I decided to combine into one PR

chalettu:
Screen shot of scenario that involves a email address that has a regex but is not required. 
We would expect the shopping cart button to be enabled if email is blank, but if its invalid make sure its disabled. This scenario also shows the email regex working as expected.  
email_validation

chalettu:
This screenshot addresses the issue with when a user requires a checkbox.  Requiring a checkbox mandates that the dialog is invalid if the user does not have the box checked.  
checkbox_fix

Loicavenel:
Great.. thanks

chriskacerguis:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1502314

PR #188, opened 2017-10-17
Don't rely on the selected binding when reloading with preselect 🦆

skateman:
This is a "hack" until we drop the jQuery calls from MIQ which can select nodes outside the angular's scope. It's ugly, it's not a thing we want but there's no better fix :cry: 

🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆🦆
https://bugzilla.redhat.com/show_bug.cgi?id=1500176


himdel:
Temporary workaround until all trees are updated, merging :)

PR #190, opened 2017-10-17
[WIP] Add support for semantic release

### LINKED ISSUE #189: Add support for Semantic release
Add support for semantic release so we can release package updates to NPM seamlessly.

chalettu:

Fixes #189

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,77 @@
+# Contributing to UI-Components
+
+ - [Commit Message Guidelines](#commit)
+
+
+## <a name="commit"></a> Git Commit Guidelines
+
+We have very precise rules over how our git commit messages can be formatted.  This leads to **more
+readable messages** that are easy to follow when looking through the **project history**.  But also,
+we use the git commit messages to **generate the AngularJS change log**.
+
+The commit message formatting can be added using a typical git workflow or through the use of a CLI
+wizard ([Commitizen](https://github.com/commitizen/cz-cli)). To use the wizard, run `yarn run commit`
+in your terminal after staging your changes in git.
+
+### Commit Message Format
+Each commit message consists of a **header**, a **body** and a **footer**.  The header has a special
+format that includes a **type**, a **scope** and a **subject**:
+
+```
+<type>(<scope>): <subject>
+<BLANK LINE>
+<body>
+<BLANK LINE>
+<footer>
+```
+
+The **header** is mandatory and the **scope** of the header is optional.
+
+Any line of the commit message cannot be longer 100 characters! This allows the message to be easier
+to read on GitHub as well as in various git tools.
+
+### Revert
+If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit.
+In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
+
+### Type
+Must be one of the following:
+
+* **feat**: A new feature
+* **fix**: A bug fix
+* **docs**: Documentation only changes
+* **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing
+  semi-colons, etc)
+* **refactor**: A code change that neither fixes a bug nor adds a feature
+* **perf**: A code change that improves performance
+* **test**: Adding missing or correcting existing tests
+* **chore**: Changes to the build process or auxiliary tools and libraries such as documentation
+  generation
+
+### Scope
+The scope could be anything specifying place of the commit change. For example `$location`,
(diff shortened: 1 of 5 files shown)


chalettu:
GH Issue https://github.com/ManageIQ/ui-components/issues/189

CONTRIBUTING.md
```diff
@@ -0,0 +1,77 @@
+# Contributing to UI-Components
+
+ - [Commit Message Guidelines](#commit)
+
+
+## <a name="commit"></a> Git Commit Guidelines
+
+We have very precise rules over how our git commit messages can be formatted.  This leads to **more
+readable messages** that are easy to follow when looking through the **project history**.  But also,
+we use the git commit messages to **generate the AngularJS change log**.
```
himdel: wrong project

himdel:
So.. as discuessed in Mahwah, I'm completely for `semantic-release` itself, but mostly against all the guidelines :).

--- 

> The **header** is mandatory and the **scope** of the header is optional.

Which makes the **type** mandatory, which is not what we agreed on.


> Any line of the commit message cannot be longer 100 characters!

No, splitting lines is a job better left for browsers, not commit message authors. We do have multiple commit messages with lines over 100 chars.


> If the commit reverts a previous commit, it should begin with `revert: `,

No reason to change git defaults.


> feat / fix / docs / style ...

We have labels for that, these should be per-PR, not per-commit.


> The scope could be anything specifying place of the commit change. For example `$location`,
 +`$browser`, `$compile`, `$rootScope`, `ngHref`, `ngClick`, `ngView`, etc...

Looks completely off topic.


> don't capitalize first letter

All our commit messages everywhere start with a capital. Let's not try to introduce guidelines which go the opposite way.

---

Apart from that, I'm not seeing any description of how to mark a commit as major/minor/patch.

(Also I don't know what happens when it encounters a commit message it doesn't like - I assume it would skip, but we need to test it :).)

chriskacerguis:
@chalettu could you please address @himdel concerns?

Fryguy:
I don't understand why the package would enforce anything with commit messages...seems completely irrelevant to ensuring the version number revving is accurate...I get that it wants to impose commit messages for CHANGELOG purposes, but we already follow a standard CHANGELOG format via https://keepachangelog.com

chalettu:
@himdel , thank you for your feedback.   So if you read the documentation for semantic-release it handles 2 major things.  It handles incrementing versions and it handles generation of changelogs to correspond with releases that it makes.  Out of the box it ships with the same contribution rules that are used by the Angular project.  That is why the CONTRIBUTING.md file has some text that is definitely out of the box.  We can change the text on that page to whatever you would like and build the rules however we like.  As for the terms that are used in the git message that get picked up on by semantic-release, I read the documentation and those are customizable.  I read through the documentation at https://github.com/semantic-release/semantic-release#plugins and we can map them to what we collectively agree makes the most sense.   Let's sit down and have a discussion around the rules we would like and come to a consensus so we can put in a PR to make the rules accommodate the way we would like the projects rules to be structured.

chalettu:
@Fryguy , this package handles two things that are both closely tied together.  It attempts to make changelog generation automatic when a release gets pushed out so it anyone can read what goes into a particular release that it is generating.  The formatting of the changelog is configurable and i am sure we  could probably get it to follow the desired format you posted(http://keepachangelog.com).  If you would like to read more about how the guts of the functionality works, the projects documentation is at https://github.com/semantic-release/semantic-release.   It enforces using specific keywords in the commit message(we can change these to whatever we would like) so it can determine how it should bump the package version number.  If someones commit message doesnt match any of the patterns we have set up, the commit still gets made in Git but we wouldn't be pushing the package to NPM automatically.  Next time someone pushes code that has a matching git commit message then it would push to NPM and version bump.  The changelog would then include all the commit messages for things that had been committed since the last release.

himdel:
Understood :) .. so.. what making this do just the bare minimum - the actual realeases, and dropping the rest?

That way... the whole `CONTRIBUTING.md` can go away and be replaced by something like:

```
This project is using semantic-release. To trigger a release, create a commit with the header starting with one of the following key-words, followed by a colon:

   * `PATCH` - release a patch version
   * `MINOR` - release a minor version
   * `MAJOR` - release a major version

For example:

"MAJOR: rewritten in PHP

(optional additional description)"
```

I don't see the need to handle changelogs from the start here, especially since we already have a way, and this would make it simple a clear for the purpose of releasing stuff.

Linting commit messages is also out of scope of this PR, so I'd just disable that.

himdel:
Apparently dead for 2 months, adding wip :)

himdel:
Closing, this will be done following https://github.com/ManageIQ/react-ui-components/pull/38

PR #193, opened 2017-10-20
Introduce targeted auto refresh logic alongside cascading logic

eclarizio:
Previously, we had to iterate through the list of refreshable fields and determine which was next to be refreshed, then queuing up requests for each field one after another.

After this change, if the dialog field has responders on it, it will use those instead of relying on the auto refresh logic. This is because right now in the ops UI, dialog fields can be edited either in the old editor or the new editor, and field responders can only be set in the new editor.

https://www.pivotaltracker.com/story/show/149808445

/cc @gmcculloug

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogUser.ts
+++ b/src/dialog-user/components/dialog-user/dialogUser.ts
@@ -12,6 +12,7 @@ export class DialogUserController extends DialogClass implements IDialogs {
   public dialogFields: any;
   public refreshableFields: Array<string>;
   public dialogValues: any;
+  public fieldAssociations: any;
   public parsedOptions: any;
   public service: any;
   public areFieldsBeingRefreshed: boolean;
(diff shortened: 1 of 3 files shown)


src/dialog-user/components/dialog-user/dialogUser.ts
```diff
@@ -94,13 +100,19 @@ export class DialogUserController extends DialogClass implements IDialogs {
    * @param value {any} This is the updated value based on the selection the user made on a particular dialog field
    */
   public updateDialogField(dialogFieldName, value) {
-    const refreshable = _.indexOf(this.refreshableFields, dialogFieldName);
     this.dialogFields[dialogFieldName].default_value = value;
     this.dialogValues[dialogFieldName] = value;
     this.saveDialogData();
-    if (refreshable > -1  && !this.areFieldsBeingRefreshed) {
+    if (this.fieldAssociations !== {}) {
```
chalettu: to make this code more readable, use lodash isEmpty  https://lodash.com/docs/4.17.4#isEmpty

PR #196, opened 2017-10-31
Allow three types of nodeIcons in the order: picture > icon > image

skateman:
For now we have two types of node images in GTL list views: fonticons and fileicons. The priority is always fonticons first, however, sometimes we have custom uploaded pictures and they should be favored even over fonticons.

My solution is to introduce the new picture type with the highest priority. This will be sent for some nodes only and it should not affect the behavior of screens where there are no custom uploaded pictures. I also dropped the `hasImage` and `hasIcon` functions and introduced `getNodeIconType` that returns with the node image type that should be displayed. Tests and examples have been also adjusted.

There is also a [ui-classic PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/2573) that updates `report_data` to provide the `picture` attribute when necessary but it's not necessary to merge before this one.

https://bugzilla.redhat.com/show_bug.cgi?id=1487056

### DIFF
--- a/src/gtl/components/data-table/dataTableComponent.ts
+++ b/src/gtl/components/data-table/dataTableComponent.ts
@@ -49,12 +49,24 @@ export class DataTableController extends DataViewClass implements IDataTableBind
   }
 
   /**
+   * Public method for retrieving what icon type should be displayed
+   * @memberof DataTableController
+   * @function getNodeIconType
+   * @param row {object} whole row with data.
+   * @param columnKey header column key.
+   * @returns {string} picture | icon | image
+   */
+  public getNodeIconType(row, columnKey) {
+    const allowedGraphics = ['picture', 'icon', 'image'];
+    if (row && row.cells) {
+      return allowedGraphics.find(item => row.cells[columnKey].hasOwnProperty(item) && !!row.cells[columnKey][item]);
+    }
+  }
+
+  /**
    * Public method for checking if column of table has an icon.
    * @memberof DataTableController
    * @function hasIcon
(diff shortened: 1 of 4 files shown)


src/gtl/components/data-table/dataTableComponent.ts
```diff
@@ -49,12 +49,30 @@ export class DataTableController extends DataViewClass implements IDataTableBind
...
+    if (row && row.cells) {
+      switch(true) {
+        case row.cells[columnKey].hasOwnProperty('picture') && !!row.cells[columnKey].picture:
+          return 'picture';
+        case row.cells[columnKey].hasOwnProperty('icon') && !!row.cells[columnKey].icon:
+          return 'icon';
+        case row.cells[columnKey].hasOwnProperty('image') && !!row.cells[columnKey].image:
+          return 'image';
+      }
+    }
```
karelhala: For some reason I don't like switch statement.
What do you say about this
```Typescript
export class DataTableController extends DataViewClass implements IDataTableBinding {
  private allowedGraphics: string[] = ['picture', 'icon', 'image'];
  //...
  private static cellHasItem(cell, item) {
   return cell.hasOwnProperty(item) && !!cell[item] && item;
  }

  public getNodeIconType(row, columnKey) {
    if (row && row.cells) {
      return this.allowedGraphics.find(item => DataTableController.cellHasItem(row.cells[columnKey], item));
    }
  }
  //...
}
```

skateman:
@miq-bot assign @martinpovolny

PR #205, opened 2017-11-21
Change ng-if to ng-show for visibility dialogField property

eclarizio:
I think this was an oversight by @chalettu, we still need invisible fields to be there in the data, but the visible flag is just so that they aren't actually shown to the end user. They still need to be rendered in HTML so that their values are stored in the data structure and sent to the API when the user submits the form.

@mkanoor This should fix the issue you found with visible fields.

@miq-bot assign @chriskacerguis


martinpovolny:
Gaprindashvili?

Loicavenel:
yes

himdel:

This was released in manageiq-ui-components 1.0.2, backported to gaprindashvili via https://github.com/ManageIQ/manageiq-ui-classic/pull/2951.

PR #206, opened 2017-11-23
Remove yarn.lock

himdel:
Follow-up to https://github.com/ManageIQ/ui-components/pull/155#issuecomment-329013722.

The idea is, same as in manageiq-ui-classic, we won't have any `yarn.lock` in the master branch.

This reduces diff spam, without affecting releases:

* yarn.lock is only ignored in gitignore, not npmignore
* running npm build runs yarn and yarn build
* => the released package will still contain `yarn.lock`.

WDYT @karelhala ?


himdel:
Ping @karelhala

himdel:
Ping @karelhala :) can you merge?

PR #208, opened 2017-11-30
Increase typescript stdlib 

romanblanco:
https://github.com/ManageIQ/ui-components/pull/77#discussion_r123964139


romanblanco:
@miq-bot assign @himdel

himdel:
LGTM, should fix the warnings about stuff coming from es6-shim, etc.

(according to https://www.typescriptlang.org/docs/handbook/compiler-options.html we may also want to add `DOM.iterable` at some point .. but not needed now)

PR #209, opened 2017-11-30
Adding ordinal numbers to the dropped elements

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1514117

PR #211, opened 2017-12-01
Add es7-shim

himdel:
Array.includes is still not supported by IE10/11 (and PhantomJS)

That breaks specs in #209.

Adding es7-shim too.

Cc @romanblanco

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "angular-ui-sortable": "^0.16.1",
     "babel-eslint": "^7.2.3",
     "es6-shim": "^0.35.3",
+    "es7-shim": "^6.0.0",
     "eslint": "~3.9.1"
   }
 }
--- a/src/vendor.ts
+++ b/src/vendor.ts
@@ -1,4 +1,5 @@
 import 'es6-shim';
+import 'es7-shim';
 import 'jquery';
 import 'jquery-ui-bundle';
 import 'bootstrap-switch';


ohadlevy:
just wondering, whats the difference between that and https://babeljs.io/docs/usage/polyfill/?

himdel:
@ohadlevy Good point.. I used es7-shim because we're already using es6-shim (in ui-service, ui-classic and here), so it sounds like the bit that's missing.

But I do hear a lot about core.js so maybe it would make sense to move from es*shim to core.js.

ohadlevy:
TBH: I'm in mixed feelings about this pr as it adds es7 syntax, while not all of the syntax has been approved yet. Can you guys work towards using a polyfill for selected items instead of adding support for all of es7 features?

himdel:
@ohadlevy What specifically are you refering to? 

As far as I know, all of the shims mentioned in https://www.npmjs.com/package/es7-shim#shims are already part of ES2017, aren't they? (I'm not seeing any of those in https://github.com/tc39/proposals .)

Also, es7-shim has nothing to do with syntax, only the library methods..

ohadlevy:
@himdel its the general note that es7 is not final yet, so it might break (even if currently its not).

ohadlevy:
@martinpovolny for the future, I think we should not merge issues that are still being discussed. (worth to ping to verify imho).

himdel:
@ohadlevy I asked Martin to merge this since we do need this for a gaprindashvili PR.

(If you want to prevent a PR from being merged, it helps to indicate so clearly by adding a negative review or something like this.. with the last message being from me here, even I though the discussion was over.)

martinpovolny:
@ohadlevy: I meant no harm here and I disagree.

We have a build tomorrow and we have a PR https://github.com/ManageIQ/ui-components/pull/209 that needs to be in the build blocked by this one.

We can have the discussion and choose a different approach even revert this trivial PR but we need this in the build and I do not want to merge stuff at the very latest moment (tomorrow).

I need to consider more context not just the fact that there's a unfinished discussion.

himdel:

This was released in manageiq-ui-components 1.0.2, backported to gaprindashvili via https://github.com/ManageIQ/manageiq-ui-classic/pull/2951.

PR #216, opened 2017-12-11
When field is removed from dialog editor refresh positions of field elements

karelhala:
### Fixes error when removing field from dialog editor
When removing dialog editor field ids are not recalculated so after clicking on edit field inside dialog group either error (when penultimate was removed) or wrong modal is shown (with wrong data).

### No UI changes.


romanblanco:
@miq-bot assign @himdel 

@himdel can you merge?

himdel:

PR #217, opened 2017-12-12
Type-neutral sort direction icon in miq-sort-items

himdel:
ui-classic's bottom toolbar is using the miq-sort-items component to choose the field to search by and to indicate the direction.

miq-sort-items is relying on pf-sort, from angular-patternfly to actually display the sort icon

But pf-sort cannot be overriden to return diffrernt icons. Also, patternfly cannot be updated to current version in gaprindashvili.

=> we need to temporarily move pf-sort out of angular-patternfly, and possibly merge the changes back in some form (like, under a `if (ctrl.config.currentField.sortType === 'neutral')` condition), after gaprindashvili.

=> Created `miq-pf-sort`, which is exactly pf-sort, except for the icons.. *and* made miq-sort-items use it.

Sort order now:

az
za

Cc @epwinchell, @karelhala 

https://bugzilla.redhat.com/show_bug.cgi?id=1501135


himdel:
@karelhala Looks like with this PR, I can reproduce the "JavaScript heap out of memory" error even on my machine :).

(doesn't matter if the last commit is `allowJs` or rename `.js` to `.ts`)


=> WIP for now, not tested

karelhala:
Looks good, this can be for now .js file and once G release is out we will move it to typescript.

himdel:

serenamarie125:
@himdel Can you confirm that the first image is of ascending sort and the second for descending sort?

himdel:
@serenamarie125 Yes, the first is a to z, the second z to a :). (Sorry, it was in the filename but github hides that.)

PR #221, opened 2017-12-14
Test for presence of 'validator_rule' instead of testing 'validator_type'

romanblanco:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1518971

Instead of testing for value of `validator_type`, test for presence of the attribute that actually has the validation rules.

/cc @chalettu


himdel:

PR #225, opened 2017-12-20
Allow the attribute to be edited in the original scope

romanblanco:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1518506

The value edited by bootstrap switch was changed in different scope, so the value of `treeSelectorIncludeDomain` in the original scope was still set on `false`.

Making the both-ways binding should fix the issue.

cc/ @h-kataria


himdel:

This was released in manageiq-ui-components 1.0.4, backported to gaprindashvili via ManageIQ/manageiq-ui-classic#3175 .

PR #228, opened 2018-01-05
Use '.name' instead of '.id' for refresh field list

eclarizio:
The problem is that when creating or editing dialogs, if you add a new field, the field is not saved before the associations are created, and so there is no id to send up. Therefore we have to send up the name and have the backend handle the association creation by name instead.

This will need to be combined with [this main repo PR](https://github.com/ManageIQ/manageiq/pull/16753) in order to fully fix the bug.

https://bugzilla.redhat.com/show_bug.cgi?id=1518942

@romanblanco @himdel Please review

### DIFF
--- a/src/dialog-editor/components/modal-field-template/check-box.html
+++ b/src/dialog-editor/components/modal-field-template/check-box.html
@@ -29,12 +29,9 @@
              switch-on-text="{{'Yes'|translate}}"
              switch-off-text="{{'No'|translate}}">
     </div>
-    <div pf-form-group pf-label="{{'Fields to refresh'|translate}}">
-      <select class="form-control" ng-model="vm.modalData.dialog_field_responders"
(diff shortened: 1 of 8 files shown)


himdel:
All those `<dialog-editor-modal-field-template...>` need a closing tag..

himdel:
Nice DRYing up :+1: 
Looks like what actually changes is..

```diff
-ng-options="dynamicField.id as dynamicField.label for dynamicField in vm.modalData.dynamicFieldList"
+ng-options="dynamicField.name as dynamicField.label for dynamicField in vm.modalData.dynamicFieldList"
```

himdel:

This was released in manageiq-ui-components 1.0.5, backported to gaprindashvili via ManageIQ/manageiq-ui-classic#3203 .

PR #229, opened 2018-01-09
Update field name only when the field (not a box) is dropped

romanblanco:
Drag&Drop of a box was causing an error of calling method for `undefined`


romanblanco:
@miq-bot add_label bug
@miq-bot assign @karelhala

romanblanco:
@karelhala updated. Thanks 👍

PR #231, opened 2018-01-10
Validate presence of values for Dropdown list and Radio button

romanblanco:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1507751

Adding another validation rule to make sure the dialog with dropdown or radio button without values is not sent.

___

One more thing here - I'd expect the same validation to be in the backend, even though normally the user shouldn't be able to send such dialog.

I've checked with @d-m-u and currently, we are setting `[[nil, "<None>"]]` as the initial value ([here](https://github.com/ManageIQ/manageiq/blob/master/app/models/dialog_field_drop_down_list.rb#L16)), and that will stay, if the dialog is submitted with no values.

Is this OK? Or should this be handled differently?

/cc @eclarizio @d-m-u

### DIFF
--- a/src/dialog-editor/components/toolbox/toolboxComponent.ts
+++ b/src/dialog-editor/components/toolbox/toolboxComponent.ts
@@ -77,7 +77,7 @@ export class ToolboxController {
         'dropdown_list',
         {
           data_type: 'string',
-          values: [],
+          values: [[1, 'One'], [2, 'Two'], [3, 'Three']],
           options: {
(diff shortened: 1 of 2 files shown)


romanblanco:
@miq-bot add_label bug
@miq-bot assign @karelhala

himdel:

This was released in manageiq-ui-components 1.0.6, backported to gaprindashvili via ManageIQ/manageiq-ui-classic#3269.

PR #234, opened 2018-01-18
Fix for inability to change a value in a dialog field's drop down list

eclarizio:
This fix removes the assignment of default_value when updating field values, as it is not necessary since the `ng-model` property on the element already binds to the correct value that gets sent when the submit button is pressed.

Here's a gif of the issue that this fix solves:
dialog-field-drop-down-broken

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530319

Thanks @syncrou for helping me debug this one, whew what an adventure!
/cc @gmcculloug

@chalett, @himdel Can you review? I tried to find a reason for the line I deleted but ultimately I don't think we need to assign the `default_value`.

### DIFF
--- a/demo/data/dialog-data.json
+++ b/demo/data/dialog-data.json
@@ -48,7 +48,7 @@
                       "display_method_options": {},
                       "required": true,
                       "required_method_options": {},
-                      "default_value": "",
+                      "default_value": "service_default",
                       "values_method_options": {},
                       "options": {
                         "protected": false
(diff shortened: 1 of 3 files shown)


eclarizio:
@miq-bot add_label gaprindashvili/yes, blocker, bug

gmcculloug:
@miq-bot assign @himdel

himdel:
LGTM, @romanblanco agrees, merging :).

himdel:

This was released in manageiq-ui-components 1.0.7, backported to gaprindashvili via ManageIQ/manageiq-ui-classic#3279.

gmcculloug:
@himdel Thanks for the updates.

@AllenBW Do we need an update to get this included in the service UI as well?

PR #238, opened 2018-01-23
DialogEditor - getDynamicFields - don't skip fields without id

himdel:
Create a new dialog, add a dynamic field, add a non-dynamic field and edit that one.
In the list of fields to refresh, the first field should be visible.

Without this, it's not, because the freshly created field has no id, and passing undefined to `getDynamicFields` would cause it to skip all the fields with undefined id.

..updated to skip undefined
..and changed to use `name` instead of `id` to keep in line with https://github.com/ManageIQ/ui-components/pull/228

(Two-part fix, this PR should go first, so that the [ui-classic PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/3302) can be tested :).)

Cc @romanblanco @eclarizio 

https://bugzilla.redhat.com/show_bug.cgi?id=1536528

### DIFF
--- a/src/dialog-editor/services/dialogEditorService.ts
+++ b/src/dialog-editor/services/dialogEditorService.ts
@@ -58,7 +58,11 @@ export default class DialogEditorService {
   public getDynamicFields(idToExclude) {
     let dynamicFields = [];
     this.forEachDialogField((field) => {
-      if (field.dynamic === true && field.id !== idToExclude) {
+      if (idToExclude && (field.id === idToExclude)) {
+        return;
+      }
+
+      if (field.dynamic === true) {
         dynamicFields.push(field);
       }
     });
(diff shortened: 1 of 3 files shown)


src/dialog-editor/services/dialogEditorService.ts
```diff
@@ -55,10 +55,14 @@ export default class DialogEditorService {
     return this.data.content[0].dialog_tabs;
   }
 
-  public getDynamicFields(idToExclude) {
+  public getDynamicFields(nameToExclude) {
     let dynamicFields = [];
     this.forEachDialogField((field) => {
-      if (field.dynamic === true && field.id !== idToExclude) {
+      if (nameToExclude && (field.name === nameToExclude)) {
```
karelhala: No need to check for `nameToExclude` since it will always be defined.

himdel:

This was released in manageiq-ui-components 1.0.8, backported to gaprindashvili via ManageIQ/manageiq-ui-classic#3304.

PR #239, opened 2018-01-24
Install dependencies before running tests

romanblanco:
To prevent ending up with errors


romanblanco:
@miq-bot assign @himdel

himdel:
Makes sense :)

PR #240, opened 2018-01-24
[WIP] BZ#1539862- Added multiselect as a dropdown option type

chalettu:
@miq-bot add_label ux/review
@miq-bot add_label bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ yarn start
 
 To run tests:
 ```
+yarn
 yarn run build-dev
 yarn run test
 ```


chalettu:
Screenshot of intended interaction for multiselect for drop downs and tag controls
multiselect_demo

chalettu:
@serenamarie125 , I wanted you to take a look and weigh in on this.  Our dialog component needs to support a tag control and a multiselect control as these are supported input types for dialogs in both SUI and Classic UI.  I am not sure what the usage pattern was for multiselect in Classic UI but this field type has not previously existed in SUI.  This control will be universal across both UI's.

chalettu:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1539862

eclarizio:
It's a lot more obvious that this seems to be the wrong component since it's multi-select, I didn't realize before with the single `ui-select` component, but I think we should just be using standard select elements with the `pf-select` property on them to make them patternfly styled.

They should look like this:
screen shot 2018-02-02 at 1 18 11 pm
screen shot 2018-02-02 at 1 18 22 pm

(Found here: https://www.patternfly.org/pattern-library/widgets/#bootstrap-select)

@chalettu Do you know if there is a specific reason we're using `ui-select` instead of the patternfly select stuff?

JPrause:
@chalettu was there further edits needed? This is for a blocker issue.

chalettu:
@JPrause I am going to create a new PR that is updated with changes.  I will link this closed PR just for history sake

PR #243, opened 2018-01-25
Introducing the quadicon component

skateman:
This component is analogous to the quadicons in the classic UI, however it has some extra features:
* Tooltips per quadrant - https://www.pivotaltracker.com/n/projects/1613907/stories/123185301
* Fonticon support - https://www.pivotaltracker.com/n/projects/1613907/stories/123185361
* Rendering from a JSON
* Shortening numbers in the text (requires [numeral](http://numeraljs.com/))

Here's an example of the format that the component understands:
```js
{
  topLeft: {
    fonticon: 'fa fa-cog',
    tooltip: 'I am a tooltip',
  },
  topRight: {
    fonticon: 'fa fa-play',
    background: 'blue',
  },
  bottomLeft: {
    text: '42',
  },
  bottomRight: {
    fonticon: 'fa fa-shield',
  },
  middle: {
    fileicon: '/assets/100/vendor-hawkular.png',
  }
}
```

And some real-life screenshots:

screenshot from 2018-01-25 15-53-59

Hopefully this will be a full replacement of the current quads we have and we can implement new features or do a redesign inside this component.

Replaces #175, resolves #222 and fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1443481
https://bugzilla.redhat.com/show_bug.cgi?id=1432491

PR #244, opened 2018-02-01
Mark numeral.js as an external dependency provided by ui-classic

skateman:
@karelhala asked @karelhala gets it, we want this as a dependency in ui-classic.

### DIFF
--- a/src/vendor.ts
+++ b/src/vendor.ts
@@ -13,3 +13,5 @@ import 'angular-dragdrop';
 import 'angular-bootstrap-switch';
 import 'ui-select';
 import 'patternfly-bootstrap-treeview';
+import * as numeral from 'numeral';
+window['numeral'] = numeral;
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -110,6 +110,7 @@ module.exports = {
   externals: {
     'angular': 'angular',
     'lodash': '_',
+    'numeral': 'numeral',
     '__': '__'
   }
 };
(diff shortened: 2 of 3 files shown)


himdel:
Makes sense.. but.. where is this being used and is *that* used in SUI?

skateman:
@himdel, we introduced this with [quadicons](https://github.com/ManageIQ/ui-components/pull/243/commits/356c37f451244589600fc79545a0ba1590efa046) for number abbreviations. I don't think SUI is already depending on it :wink:

himdel:
Perfect.. so now only the failing specs.. :)

```
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  ReferenceError: Can't find variable: numeral
  at dist/js/ui-components.js:9
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  ReferenceError: Can't find variable: numeral
  at src/common/filters/abbrNumberFilter.spec.ts:74
```

.. looks like numeral is missing from `package.json`? :)

skateman:
@karelhala need some help here

karelhala:
@skateman update `src/vendor.ts` to add `numeral` as global package. Otherwise webpack will try to include it to the final build, but tests are trying to find them at window.

```diff
diff --git a/src/vendor.ts b/src/vendor.ts
index 30ac03d..8c84839 100644
--- a/src/vendor.ts
+++ b/src/vendor.ts
@@ -13,4 +13,6 @@ import 'angular-dragdrop';
 import 'angular-bootstrap-switch';
 import 'ui-select';
 import 'patternfly-bootstrap-treeview';
-import 'numeral';
+import * as numeral from 'numeral';
+
+window['numeral'] = numeral;
```

skateman:
@himdel now it should be fine!

PR #245, opened 2018-02-01
Display the dialog's JSON in the dialog editor demo page

skateman:
Theoretically if someone would want to replace an old dialog with dialogPlayer + JSON, there is no easy way to create the JSON. This one-liner is to fix this problem!

screenshot from 2018-02-01 16-44-53

@karelhala @himdel @romanblanco


himdel:
(Demo-only change, no gaprindsahvili.)

PR #246, opened 2018-02-02
BZ#1539871 - Fix for add to Shopping Cart being disabled

chalettu:
@miq-bot add_label bug

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogUser.ts
+++ b/src/dialog-user/components/dialog-user/dialogUser.ts
@@ -214,6 +214,7 @@ export class DialogUserController extends DialogClass implements IDialogs {
           this.updateTargetedFieldsFrom(field);
         } else if (this.refreshRequestCount === 0) {
           this.areFieldsBeingRefreshed = false;
+          this.saveDialogData();
         }
 
         resolve(data);


chalettu:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1539871

chalettu:
Screenshot of it ensuring that the add to cart enables on the last field after it has been updated.  
dynamic field refresh

himdel:
Makes sense, `saveDialogData` is being called twice, but the second call is needed to pass validation, and the first one is presumably still relevant for `updateTargetedFieldsFrom`, right? :)

himdel:
.. but.. the first call is followed by `$scope.$apply()`, presumably because this is using `Promise`, not `$q` .. won't we need that apply after the second call as well?

chalettu:
Yes, both are still needed.  The first one deals with responding to the data change that came back from refreshing the dialog field.  The second one is right after we flip a bit flag saying no more fields are left to refresh.  This then goes through and revalidates all fields so that the form becomes valid.  We built a mode that fails validation if you try to validate while dialog fields are refreshing.  @himdel .  In testing we haven't seen that we need that scope apply .  I believe its because all the data that changed on the dialog field was called in that scope.apply.  The data about validation is one that is passed up to the parent component so scope.apply shouldnt need to apply in that scenario.

himdel:
@chalettu Agreed, clicked through in the ui, and seems like nothing else is needed :).

himdel:
Backported to `gaprindashvili` branch...

```
commit 322071259153db44255c8c962e26eb741c10e44c (HEAD -> gaprindashvili)
Date:   Tue Feb 6 17:09:07 2018 +0100

    Merge pull request #246 from chalettu/add_to_cart_disabled
    
    BZ#1539871 - Fix for add to Shopping Cart being disabled
    
    (cherry picked from commit 5fe8fd5cf5c9634b51f9f52b8890d490abd7e0f3)
```

himdel:
Released in manageiq-ui-components 1.0.9

PR #247, opened 2018-02-05
Ensure default value for radio button is rendered properly

eclarizio:
There was a bug in the way that the default value was being sent to the API in that it was trying to send the full value (so for example ["1", "one"] instead of just "1"). The way that the `ng-checked` property works was also not actually as intended, all that needs to be done is to ensure that `ng-value` and the `ng-model` are equivalent and angular will handle selecting the correct value. The `ng-checked` property is for if you want to send up different data than what is contained in the `ng-value` property. This fixed the actual correct default value showing up in the editor, as seen in the below screenshots:

screen shot 2018-02-05 at 12 11 47 pm
screen shot 2018-02-05 at 12 12 43 pm

The other issue was that the "Nothing selected" value was always selected from the dropdown list when editing the default value for radio buttons, and this has been fixed to show which value is actually the default value:

screen shot 2018-02-05 at 12 11 55 pm
screen shot 2018-02-05 at 12 12 48 pm

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1541988

@miq-bot assign @himdel

### DIFF
--- a/src/dialog-editor/components/field/field.html
+++ b/src/dialog-editor/components/field/field.html
@@ -66,8 +66,7 @@
         <input type="radio"
                name="{{vm.fieldData.name}}"
                ng-model="vm.fieldData.default_value"
-               ng-checked="vm.fieldData.default_value === option"
-               ng-value="option">
+               ng-value="option[0]">
         {{ option[1] }}
       </label>
     </span>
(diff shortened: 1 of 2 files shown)


himdel:
Backported to `gaprindashvili` branch:

```
commit d2ef03a74f6f4e2c8b26e1d4f5776609a6813d9f (HEAD -> gaprindashvili)
Date:   Tue Feb 6 15:54:24 2018 +0100

    Merge pull request #247 from eclarizio/BZ1541988
    
    Ensure default value for radio button is rendered properly
    
    (cherry picked from commit e9d41ef2d585b923b8f06dbae39856cb1b3b0932)
```

himdel:
Released in manageiq-ui-components 1.0.9

PR #254, opened 2018-02-13
Update Bower dev CI to install dependencies before releasing

karelhala:
### Fixes problem with bower releases
When releasing new version to bower it does not install dependencies which causes error, this PR fixes such issue and before releasing new version to bower it installs required dependencies.

### DIFF
--- a/ci/bower-dev.sh
+++ b/ci/bower-dev.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -x
 HASH=$(git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


karelhala:
@himdel this should fix our build into bower.
@skateman please take a look at this.

himdel:
This .. will not affect the automated build at all, our `.travis.yml` already has `yarn: true` so `yarn install` happens before the script gets run.

But might as well add it in case somebody runs this manually :).

PR #255, opened 2018-02-14
Add missing code to make the dialog editor working in the demo

skateman:
Some copy-pasted code from ui-classic's `dialog_editor_controller.js` and `editor.html` that will make the modals work in the demo.


romanblanco:
@miq-bot add_label gaprindashvili/no
@miq-bot add_label bug
@miq-bot assign @romanblanco

PR #256, opened 2018-02-15
Drop the unnecessary visible binding from dialogEditorModal component

skateman:
@romanblanco this binding is never being used anywhere...


romanblanco:
@miq-bot add_label refactoring
@miq-bot add_label cleanup
@miq-bot assign @karelhala

PR #259, opened 2018-02-16
Pluralize the box component to boxes in DialogEditor

skateman:
The box component was not representing a single box, but a list of them...so renaming it.

@miq-bot assign @romanblanco


skateman:
This is wrong, closing...

PR #263, opened 2018-02-19
Add quadicon into report data, small tile and big tile

karelhala:
### Introduces quadicons as part of report_data component
When rendering report_data tile (small/big) show quadicon instead of using HTML when possible.

### UI changes
screenshot from 2018-02-19 17-04-37

### DIFF
--- a/demo/data/data-table.json
+++ b/demo/data/data-table.json
@@ -53,7 +53,26 @@
           {
             "text": "server-one"
           }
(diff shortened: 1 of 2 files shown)


himdel:
Ok, so, `item.quad` is a json definition, whereas `item.quadicon` is the plain old html.

And now we support both.. sounds good :)

PR #227, opened 2018-01-03
Fix assumption that all field.values are objects

eclarizio:
This got changed back in https://github.com/ManageIQ/ui-components/pull/223, but it makes it so that when you try to edit anything other than drop down or radio button fields (the ones that use objects for their values), it attempts to call `.filter` on a string and ends up blowing up silently in JS with you unable to edit your fields.

@romanblanco @martinpovolny please review

@miq-bot add_label gaprindashvili/yes

PR #264, opened 2018-02-20
Move default value compensation to loadDialog in the DialogEditorService

skateman:
Introduced in #227, but definitely in a wrong place...moving it closer to the data loading.

\cc @eclarizio @romanblanco


skateman:
@miq-bot assign @romanblanco

romanblanco:
@miq-bot assign @karelhala

PR #265, opened 2018-02-20
Updated support for sorting of drop downs

chalettu:
@miq-bot add_label bug
This PR also addresses an issue where "None" option should default to staying at the top of a dropdown list of options.
Fix for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1531677

### DIFF
--- a/src/dialog-user/services/dialogData.spec.ts
+++ b/src/dialog-user/services/dialogData.spec.ts
@@ -127,7 +127,7 @@ describe('DialogDataService test', () => {
         [5, 'Test2'],
         [2, 'Test5']
       ],
-      options: { sort_by: 'value', sort_order: 'descending' }
+      options: { sort_by: 'value', sort_order: 'descending', data_type: 'integer' }
     };
     const testSorted = dialogData.updateFieldSortOrder(testDropDown);
     const expectedResult = [[5, 'Test2'], [2, 'Test5'], [0, 'Test']];
--- a/src/dialog-user/services/dialogData.spec.ts
+++ b/src/dialog-user/services/dialogData.spec.ts
@@ -144,4 +144,17 @@ describe('DialogDataService test', () => {
     const expectedSortedResult = [[5, 'C'], [0, 'B'], [2, 'A']];
     expect(testSortedDescription).toEqual(expectedSortedResult);
   });
+  it('should allow a numeric Description field to be sorted in a dropdown', () => {
+    const testDropDownDescription = {
+      values: [
+        [0, '0'],
+        [5, '5'],
+        [2, '2']
+      ],
+      options: { sort_by: 'description', sort_order: 'descending' }
+    };
+    const testSortedDescription = dialogData.updateFieldSortOrder(testDropDownDescription);
+    const expectedSortedResult = [[5, '5'], [2, '2'], [0, '0']];
+    expect(testSortedDescription).toEqual(expectedSortedResult);
+  });
 });
--- a/src/dialog-user/services/dialogData.spec.ts
+++ b/src/dialog-user/services/dialogData.spec.ts
@@ -147,14 +147,14 @@ describe('DialogDataService test', () => {
   it('should allow a numeric Description field to be sorted in a dropdown', () => {
     const testDropDownDescription = {
       values: [
-        [0, '0'],
-        [5, '5'],
-        [2, '2']
+        ['zero', '0'],
+        ['five', '5'],
+        ['two', '2']
       ],
       options: { sort_by: 'description', sort_order: 'descending' }
     };
     const testSortedDescription = dialogData.updateFieldSortOrder(testDropDownDescription);
-    const expectedSortedResult = [[5, '5'], [2, '2'], [0, '0']];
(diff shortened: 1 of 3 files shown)


chalettu:
Screenshot of Dropdown sorted Asc with value set to Int and sorting on description field. 
sortasc

chalettu:
Screenshot of a dropdown thats values are a string and sorted by description.  Notice the None staying on top
stringsortasc

chalettu:
Screenshot of radio button groups sorted desc and asc.  
radio

src/dialog-user/services/dialogData.ts
```diff
@@ -48,8 +46,26 @@ export default class DialogDataService {
...
+    const VALUE = 0;
+    const sortBy = (data.options.sort_by === 'value' ? SORT_VALUE : SORT_DESCRIPTION);
+    let tempValues = data.values;
+    let defaultDropdownField = [];
+    // The following if deals with a empty default option if it exists
+    if (data.data_type === 'integer' && _.isNaN(tempValues[FIRST_OPTION][VALUE])) {
+      defaultDropdownField = tempValues.shift();
+    } else if(_.isNull(tempValues[FIRST_OPTION][VALUE])) {
+      defaultDropdownField = tempValues.shift();
+    }
```
eclarizio: Should we just combine this `else if` and the `if` with an `||` since both lines 58 and 60 do the same thing?

chalettu: I thought about that when i reviewed my own code.  I think I will try that but will put the or on a separate line

src/dialog-user/services/dialogData.ts
```diff
@@ -25,11 +25,9 @@ export default class DialogDataService {
         if (option[0] === String(field.default_value)) {
           field.selected = option;
         }
-        if (field.data_type === 'integer') {
-          dropDownValues.push([parseInt(option[0], 10), option[1]]);
-        } else {
-          dropDownValues.push(option);
-        }
+        const value = (field.data_type === 'integer' ? parseInt(option[0], 10) : option[0]);
+        const description = (parseInt(option[1], 10) ? parseInt(option[1], 10) : option[1]);
```
himdel: Is this `description` used only for sorting?

Because if not, this breaks on any description which starts with a number.  `"3 horses"` will become just `"3"`

chalettu: Description is only used for sorting the drop down list.

src/dialog-user/services/dialogData.spec.ts
```diff
@@ -144,4 +144,17 @@ describe('DialogDataService test', () => {
     const expectedSortedResult = [[5, 'C'], [0, 'B'], [2, 'A']];
     expect(testSortedDescription).toEqual(expectedSortedResult);
   });
+  it('should allow a numeric Description field to be sorted in a dropdown', () => {
+    const testDropDownDescription = {
+      values: [
+        [0, '0'],
```
himdel: Also, can you change the keys and values to something different?

It is impossible to tell if this really works or if it sorts by value instead :).

himdel: Oh, and also without an `11` or something like that, it doesn't actually tell us if those are sorted as numbers or as strings..

eclarizio: Also, we need to set the `data.data_type` to `integer` which is part of the original problem that javascript was sorting 1, 11, and 2 as if they were strings.

src/dialog-user/services/dialogData.ts
himdel: Oh.. and maybe an `_.isNaN`? 

Otherwise 0 will still get sorted last..

chalettu: ha, you beat me to it.  I just added isNan in just now.

src/dialog-user/services/dialogData.spec.ts
chalettu: Actually that wasn't the problem this PR addresses.  This deals with a use case where a user would decide to actually use numbers in the description field and then decide they want to sort the dropdown by the description.

chalettu: @eclarizio , I went ahead and updated one of the test cases to actually specify the data type.  It wasn't directly related to this test but I added it in for clarity sake.

himdel:
Released in 1.1.7 (master only)

himdel:
Backported to `gaprindashvili`:

```
commit b4dbb3407b4a7d64a02b6d38d4ec1f6333bc2889 (HEAD -> gaprindashvili, upstream/gaprindashvili)
Date:   Wed Feb 21 18:07:03 2018 +0000

    Merge pull request #265 from chalettu/dialog-sorting-dropdowns
    
    Updated support for sorting of drop downs
    
    (cherry picked from commit fa8af254089eb5e6e9425a2fbfdcacdfe2886f67)
```

JPrause:
Awesome! Thanks @chalettu @eclarizio @himdel

himdel:
Released in 1.0.16 (gaprindashvili)

PR #266, opened 2018-03-08
Add ability to live-search dialog drop down fields

eclarizio:
When we switched over to the new dialog-user component, we forgot to include the live-search ability that was on the older bootstrap drop downs. This will put those back into place, and adds the `data-tokens` property to the options so for example if someone has a drop down that submits the value of `3` but the option for selecting is `"Three"`, then they can search by typing either `3` or `three`.

See screenshots:
screen shot 2018-03-08 at 11 00 52 am
screen shot 2018-03-08 at 11 00 43 am

https://bugzilla.redhat.com/show_bug.cgi?id=1553347

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -50,20 +50,43 @@
       </div>
 
       <span ng-switch-when="DialogFieldDropDownList">
-        <!-- Dropdown field where a single value is expected -->
+        <!-- Dropdown field where a single value is expected - PF 3 compatible-->
         <select pf-select
-                ng-if="!vm.dialogField.options.force_multi_value"
+                data-live-search="true"
+                ng-if="!vm.dialogField.options.force_multi_value && vm.patternflyVersion === 3"
                 ng-model="vm.dialogField.default_value"
                 ng-blur="vm.validateField()"
                 ng-change="vm.changesHappened()"
                 ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
                 class="form-control"
                 id="{{ vm.dialogField.name }}">
-          <option ng-repeat="value in vm.dialogField.values" value="{{value[0]}}">{{value[1]}}</option>
+          <option ng-repeat="value in vm.dialogField.values"
+                  data-tokens="{{value[0]}} {{value[1]}}"
+                  value="{{value[0]}}">
+            {{value[1]}}
+          </option>
+        </select>
+
+        <!-- Dropdown field where a single value is expected - PF 4 compatible-->
+        <select pf-bootstrap-select
+                data-live-search="true"
+                ng-if="!vm.dialogField.options.force_multi_value && vm.patternflyVersion === 4"
+                ng-model="vm.dialogField.default_value"
+                ng-blur="vm.validateField()"
+                ng-change="vm.changesHappened()"
+                ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
+                class="form-control"
+                id="{{ vm.dialogField.name }}">
+          <option ng-repeat="value in vm.dialogField.values"
+                  data-tokens="{{value[0]}} {{value[1]}}"
+                  value="{{value[0]}}">
+            {{value[1]}}
+          </option>
         </select>
 
         <!-- PF 3 compatible multiselect -->
         <select pf-select multiple
+                data-live-search="true"
                 ng-if="vm.dialogField.options.force_multi_value && vm.patternflyVersion === 3"
                 ng-init="vm.convertValuesToArray()"
                 ng-model="vm.dialogField.default_value"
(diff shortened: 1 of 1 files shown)


eclarizio:

@miq-bot assign @himdel 

chalettu:
@eclarizio this looks pretty good but one question.  What field is the livesearch searching on.  It looks like its searching the value.  Is this the case?  I think if that is the case it makes more sense to search the text of the dropdown which i think is value[1] not 0.

eclarizio:
@chalettu Oops, it should be doing both. Just updated to include `value[1]` as well in the `data-tokens` so that it works for both.

chalettu:
Looks good to me!

himdel:
Code looks good, tested in ui-classic :+1: 

Just want to make sure single select is working even in SUI, because we're not using pf4 bootstrap select except for multiselect...

himdel:
Ok, so, this is currently completely broken in SUI, at least for me.

Accessing the screen `/catalogs/1` gives me plain old selects ... @chalettu does this happen for you too?

If I try to fix this...

```diff
diff --git a/client/app.js b/client/app.js
index 21df751f..a3dcabad 100644
--- a/client/app.js
+++ b/client/app.js
@@ -46,6 +46,7 @@ require('angular-dragdrop')
 require('angular-drag-and-drop-lists/angular-drag-and-drop-lists')
 require('patternfly/dist/js/patternfly')
 require('angular-patternfly/dist/angular-patternfly')
+require('bootstrap-select/js/bootstrap-select')
 
 // Libraries: misc
 require('ngprogress/build/ngprogress')
diff --git a/package.json b/package.json
index dbbdfd71..52c9fd2f 100644
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "bootstrap": "3.1.1",
     "bootstrap-combobox": "1.0.2",
     "bootstrap-datepicker": "1.7.1",
+    "bootstrap-select": "^1.12.4",
     "bootstrap-switch": "3.3.4",
     "c3": "0.4.21",
     "clean-webpack-plugin": "0.1.18",
```

Then the first dialog indeed doesn't do search, becuase we'd need `pf-bootstrap-select` there, not `pf-select`.

@eclarizio , @gmcculloug do we merge this now so that this is working at least in ui-classic, or should we wait until the search works in both?

eclarizio:
@himdel Wait, this PR broke the selects so that SUI only has plain selects? I didn't change the `pf-select` attribute on any of the fields, that's odd.

I would say *if* the SUI currently already only has regular selects without this PR then I would think we should just merge this now and fix for service UI later.

himdel:
@eclarizio there's a current SUI bug where neither pf-select nor pf-bootstrap-select work. That bug is not caused by your PR and is present only in master, not gaprindashvili. @AllenBW is looking into that.

But the problem in *this* PR is that `pf-select` only knows about `data-live-search` in pf3. In pf4, you'd need `pf-bootstrap-select` for that, so you'd need the same kind of pf3/pf4 condition even for singleselect.

eclarizio:
Updated to include a PF3/4 version of the single select

himdel:
Thanks, verified this is still working in Ops.
And *almost* sure this is doing the right thing in SUI (with the hacks mentioned above).

Merging once green..

himdel:
Released in `@manageiq/ui-components` 1.1.8.


Backported to gaprindashvili:

```
commit 9c5dd6da9e8e8d1bd929749c9cfe078524b1a04e (HEAD -> gaprindashvili)
Date:   Tue Mar 13 15:42:01 2018 +0000

    Merge pull request #266 from eclarizio/BZ1553347
    
    Add ability to live-search dialog drop down fields
    
    (cherry picked from commit c14394800b7942f49c868ebedd6e7bbb200513f7)
```


...

himdel:
.. and released in `@manageiq/ui-components` 1.0.18 and `manageiq-ui-components` 1.0.18.

AllenBW:
screen shot 2018-03-13 at 1 34 33 pm

☝️ current sui master looks like we're not using the pf-4 version of the select... for whatever reason its rendering [the pf3 version](https://github.com/ManageIQ/ui-components/blob/master/src/dialog-user/components/dialog-user/dialogField.html#L54)

but I coulda sworn https://github.com/ManageIQ/manageiq-ui-service/pull/1379 was made to fix this... any thoughts @chalettu ?

AllenBW:
On master `.bootstrap-select` assets are available, looks like we're not getting the correct dialog to show up though  👇 

screen shot 2018-03-13 at 3 29 11 pm

AllenBW:
Ongoing saga 😩 jumping master to ui-c 1.1.8, we now see the correct (pf-4 select) option rendered, but still missing style... 🤔

himdel:
This should now be working everywhere, thanks @AllenBW  :)

PR #267, opened 2018-03-09
When no checkboxes are visible in GTL hide select all

karelhala:
### If no checkboxes in GTL remove select all
When user is in GTL view screen and no checkboxes are available select all is still present, this is wrong. This PR fixes such issue by adding new method in `src/gtl/interfaces/abstractDataViewClass.ts` and setting one way data bounded `hasCheckboxes` in `src/gtl/components/pagination/paginationComponent.ts`

#### Before
screenshot from 2018-03-09 14-07-07

#### After
screenshot from 2018-03-09 14-07-14

### Merge before core
This PR is multiple PR work, it needs to be merged and new version released before merging TODO ui-classic.

### DIFF
--- a/src/gtl/interfaces/abstractDataViewClass.ts
+++ b/src/gtl/interfaces/abstractDataViewClass.ts
@@ -81,6 +81,22 @@ export abstract class DataViewClass implements IDataTableBinding {
       `${start} - ${end} of ${total}`;
   }
 
+  /**
+   * Helper method to count all checkboxes in rows data.
+   * Checkboxes are stored under each row's cells.
+   */
+  public countCheckboxes() {
+    return this.rows.reduce(
+      (curr: number, next) => {
+        if (next.cells) {
+          curr += next.cells.filter(oneCell => oneCell && oneCell.is_checkbox).length;
+        }
+        return curr;
+      },
+      0
+    );
+  }
+
   public onItemButtonClick(item: any, $event: any) {
     $event.stopPropagation();
     if (item.hasOwnProperty('onclick')) {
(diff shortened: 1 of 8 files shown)


src/gtl/interfaces/abstractDataViewClass.ts
```diff
@@ -81,6 +81,17 @@ export abstract class DataViewClass implements IDataTableBinding {
       `${start} - ${end} of ${total}`;
   }
 
+  /**
+   * Helper method to count all checkboxes in rows data.
+   * Checkboxes are stored under each row's cells.
+   */
+  public countCheckboxes() {
+    return this.rows.reduce(
+      (curr: number, next) => next.cells && next.cells.filter(oneCell => oneCell && oneCell.is_checkbox).length + curr,
```
himdel: Should this be `curr + (next.cells && ...)`?

Because, now any row with no cells will reset the counter to `undefined` or `NaN`.

karelhala: Good catch. However this is even better and more robust
```javascript
return this.rows.reduce(
  (curr: number, next) => {
    if (next.cells) {
      curr += next.cells.filter(oneCell => oneCell && oneCell.is_checkbox).length
    }
    return curr;
  },
  0
);
```

karelhala: And tests written to cower this isse as well.

karelhala:
@himdel Can we merge this and release new version? It's blocker for https://bugzilla.redhat.com/show_bug.cgi?id=1553157

himdel:
Seems to work, merging :)

The BZ is not for gaprindashvili, so I assume we don't need a new version?

PR #268, opened 2018-03-09
Do not open browser when running tests

karelhala:
### No browser window for tests
When running tests new window is opened in browser which is then unresposnive, because server terminates itself once tests are done. This PR fixes such issue by changeing webpack config to allow passing variables


karelhala:
@skateman you'll like this one.

PR #270, opened 2018-03-15
Fix for drop-down selection on dialogs while ordering service catalog having to select an item twice

eclarizio:
The main problem here is that we need to use `ng-repeat` to list out the options for a select. This, however, only supports strings for the value field, so when the drop down has a `data_type` of integer, we were coercing the default value into an integer, which would cause a mismatch and nothing would be selected. I'm not exactly sure why the weird double select was happening with this, but removing the integer coercion fixes the issue.

Here's what it looks like before (note that nothing is selected by default because it's trying to match `2` with `'2'`):
drop_down_not_working

Here's what it looks like after:
drop_down_working

I've made sure that on the back end, when we actually do submit the dialog, that the request gets the correct type. It is being coerced on the back end via https://github.com/ManageIQ/manageiq/blob/master/app/models/dialog_field.rb#L108 and therefore I believe the coercion in javascript here can be safely removed.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552041

@chalettu Please review.
@miq-bot assign @himdel 

### DIFF
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -99,9 +99,6 @@ export default class DialogDataService {
     if (data.default_value) {
       defaultValue = data.default_value;
     }
-    if (data.data_type === 'integer') {
-      defaultValue = parseInt(data.default_value, 10);
-    }
 
     return defaultValue;
   }


AllenBW:
Also fixes almost fixes https://bugzilla.redhat.com/show_bug.cgi?id=1557482 (sui will need to consume version bump for total fix)

eclarizio:
@chalettu Yeah, in the screenshot, the first drop down is a dynamic one, and the second is a static one.

himdel:
Backported to gaprindashvili:

```
commit cb560ad515df4dfde4c74e674bfe3abca7b97c01 (HEAD -> gaprindashvili)
Date:   Tue Mar 20 19:35:00 2018 +0000

    Merge pull request #270 from eclarizio/BZ1552041
    
    Fix for drop-down selection on dialogs while ordering service catalog having to select an item twice
    
    (cherry picked from commit f511e15d6a808cfacac832e26c6296e7d189c41f)
```

himdel:
Released in 1.1.9 and 1.0.19

PR #272, opened 2018-03-20
Ensure 'null' default values do not get coerced into 'NaN' for integer type drop downs

eclarizio:
@AllenBW can probably give a bit more context on this, but basically if you do not set a default value for an integer based drop down, it will default to `null`. Well, this gets converted into "NaN", which does not match `null` as a default value and therefore doesn't end up actually selecting the first value which should be `"<None>"` but instead just shows "Nothing Selected".

Before, the string drop downs work fine with no default, and the integer ones have the "Nothing Selected" text:
screen shot 2018-03-20 at 2 48 30 pm

After, all of them behave the same:
screen shot 2018-03-20 at 2 48 42 pm

https://bugzilla.redhat.com/show_bug.cgi?id=1557482

@chalettu Can you please review?

@miq-bot assign @himdel 

I'm not tracking the BZ quite as closely as @AllenBW, so Allen, if I got the wrong labels please correct me 👍 

### DIFF
--- a/src/dialog-user/services/dialogData.spec.ts
+++ b/src/dialog-user/services/dialogData.spec.ts
@@ -52,6 +52,26 @@ describe('DialogDataService test', () => {
     expect(configuredField.errorMessage).toBeDefined();
   });
 
+  describe('#setupField', () => {
+    describe('when the field is a drop down list', () => {
+      describe('when the field is an integer type', () => {
+        describe('when the field has no default value', () => {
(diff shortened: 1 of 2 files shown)


dclarizio:
Merging on @himdel's behalf since he is on PTO.

chalettu:
Changes look good.

himdel:
(Released in 1.0.20, and 1.1.10)

PR #204, opened 2017-11-20
Fixed issue with Dialog field validation passing when no values selected

chalettu:
@miq-bot add_label bug
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1512398

PR #278, opened 2018-04-13
Fix for Catalog Item with Tag Control element cannot be ordered

eclarizio:
Before, we were using lodash's `_.isEmpty` method to determine if something was selected or not, but `_.isEmpty` does not behave the Ruby version we know and love. It only determines if an object is empty, not a variable. So, the problem was that the tag control select was passing in a category id, which was a number, and `_.isEmpty(1234)` returns true. Therefore validation was failing 😢.

This changes the logic so that when a single value is to be selected, we check to see if it's a number. If it's 0 (which is the "choose" value), then we still need to select something, otherwise we should be ok. If it somehow gets set to null, we should still also be failing validation.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1567108

@miq-bot assign @himdel 

@himdel It's not currently marked as a blocker but will likely be a blocker for 5.9.3, just so you know.
@chalettu Can you review please?

/cc @gmcculloug

### DIFF
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -124,6 +124,11 @@ export default class DialogDataService {
       if (field.type === 'DialogFieldCheckBox' && fieldValue === 'f') {
         validation.isValid = false;
         validation.message = __('This field is required');
+      } else if (field.type === 'DialogFieldTagControl') {
+        if (this.isInvalidTagControl(field.options.force_single_value, fieldValue)) {
+          validation.isValid = false;
+          validation.message = __('This field is required');
+        }
       } else if (_.isEmpty(fieldValue)) {
         validation.isValid = false;
         validation.message = __('This field is required');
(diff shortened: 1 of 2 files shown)


src/dialog-user/services/dialogData.ts
```diff
@@ -143,4 +148,31 @@ export default class DialogDataService {
...
+   **/
+  private isInvalidTagControl(forceSingleValue, fieldValue) {
+    let invalid = false;
+
+    if (forceSingleValue) {
+      if (_.isNumber(fieldValue)) {
+        if (fieldValue === 0) {
+          invalid = true;
+        }
+      } else if (_.isEmpty(fieldValue)) {
```
chalettu: I think this can be made a little more concise by removing it from here and removing the else below and just changing the else to an if.

src/dialog-user/services/dialogData.ts
```diff
@@ -143,4 +148,31 @@ export default class DialogDataService {
...
+        invalid = true;
+      }
+    } else {
```
chalettu: Remove the else but keep the if statement that is inside this.  Same logic, just cuts down on a few lines

eclarizio: It's not the same logic though, because `_.isEmpty` will return true when the value that gets passed in is a number, which is wrong.

himdel:
@eclarizio So, the default "bad" branch used to be `fieldValue === ""` before it got changed to `_.isEmpty(fieldValue)` in https://github.com/ManageIQ/ui-components/pull/204. 

If `isEmpty` is always wrong for non-objects, shouldn't we also change the else condition to trigger on `!fieldValue || (_.isObject(fieldValue) && _.isEmpty(fieldValue))`?

eclarizio:
@himdel Good point, I think the other scenario where it could fail then is when the field is required and the `data_type` of the field is set to `integer`, because then `_.isEmpty` would return true. I'll test today and make sure, but you're right in that I don't know if we want to hold up this PR for that.

himdel:
> I'll test today and make sure, but you're right in that I don't know if we want to hold up this PR for that.

@eclarizio As long as you're on it :).

OK, so looks like we can merge? :)

eclarizio:
I think so. @chalettu You good with this?

himdel:
Released in 1.1.11.

Backported to gaprindashvili:

```
commit d038b37063b03cf7bd2d1c5a10cc1b19b3fb7c14 (HEAD -> gaprindashvili)
Date:   Mon Apr 16 17:15:40 2018 +0200

    Merge pull request #278 from eclarizio/BZ1567108
    
    Fix for Catalog Item with Tag Control element cannot be ordered
    
    (cherry picked from commit c1319ef1a90f6e9cfc4dc093d6e62f94aad77a23)
```

Released in 1.0.22.

PR #281, opened 2018-04-19
Create new kebab component and toolbar click to merge toolbar custom buttons together

karelhala:
### Combines multiple custom buttons to kebab menu
When on screen with toolbar and many custom buttons These buttons should be combined together into kebab menu. If such button is group of buttons this feature has to be consistent and hovering above such button in kebab new drop down will be visible (preferably drop right).

This PR implements such issue by introducing one new component and one directive.
* New directive is for MiQ action buttons so we don't have to repeat same settings all over again.
* New component implements two level dropdown menu with kebab icon and whenever more than **3** custom buttons are present in toolbar they will be squashed together into kebab menu.

#### Before
screenshot from 2018-04-20 14-08-25

#### After
screenshot from 2018-04-18 16-48-52

### Requirements
* [x] write new components
* [x] write tests

### BZ
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552064

### DIFF
--- a/src/toolbar/components/toolbar-menu/toolbarComponent.ts
+++ b/src/toolbar/components/toolbar-menu/toolbarComponent.ts
@@ -1,6 +1,9 @@
 import {IToolbarItem} from '../../interfaces/toolbar';
 import {ToolbarType} from '../../interfaces/toolbarType';
 import * as _ from 'lodash';
+
+const CUSTOM_ID = 'custom_';
+
 /**
  * @memberof miqStaticAssets
  * @ngdoc controller
@@ -105,6 +108,10 @@ export class ToolbarController {
     return ToolbarType.BUTTON;
   }
 
+  public getToolbarKebabType(): string {
+    return ToolbarType.KEBAB;
+  }
+
   /**
    * Helper method for getting string value of {@link ToolbarType.CUSTOM}
    * @memberof ToolbarController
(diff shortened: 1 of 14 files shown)


karelhala:
@martinpovolny this should address https://bugzilla.redhat.com/show_bug.cgi?id=1552064, do you think I should tag UX as well?

@himdel can you take the code review?

src/toolbar/components/toolbar-menu/toolbarComponent.ts
```diff
@@ -119,6 +126,25 @@ export class ToolbarController {
...
 
+  public collapseButtons() {
+    let customButtonsIndex;
+    if (this.toolbarItems) {
+      if (this.toolbarViews && this.toolbarViews.length > 0) {
+        customButtonsIndex = this.toolbarItems.length - 2;
+      } else {
+        customButtonsIndex = this.toolbarItems.length - 1;
+      }
+        this.toolbarItems[customButtonsIndex] =
```
himdel: surprising indent

himdel:
LGTM, except for that one weirdly indented line (was it supposed to be inside that `else` block?) :)

src/toolbar/components/toolbar-menu/toolbarComponent.ts
karelhala: The line was too long so I split the assignment to new line. But this might look better
```javascript
this.toolbarItems[customButtonsIndex] = ToolbarController
  .createKebabFromItems(this.toolbarItems[customButtonsIndex]);
```

himdel: Oh, the problem is not what happens after that `=`.

It's where the line starts..

PR #284, opened 2018-04-20
 Adjust the opacity of custom colored icons when they are disabled

skateman:
If a custom color is defined for a toolbar button's icon, the `.disabled` property is not being respected on it. I created a new `adjustColor` filter that has an extra boolean argument. If the boolean is true or the input is unsed, the filter just returns the input, otherwise it converts the hex color into an `rgba(x, y, z, 0.5)` value.

**Before:**
screenshot from 2018-04-20 15-07-03

**After:**
screenshot from 2018-04-20 14-44-42

@miq-bot add_reviewer @epwinchell 
@miq-bot assign @karelhala 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1501114

### DIFF
--- a/src/common/filters/abbrNumberFilter.spec.ts
+++ b/src/common/filters/abbrNumberFilter.spec.ts
@@ -13,5 +13,5 @@ describe('Filter: abbrNumber', function() {
   it('returns with the same text', function() {
     expect(filter('text')).toBe('text');
     expect(filter('123text')).toBe('123text');
-  })
+  });
 });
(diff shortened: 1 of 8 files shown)


skateman:
@karelhala I'm struggling with the tests a litte bit, it seems the filter is not being included in them. Can you please help me?

skateman:
@martinpovolny this is a missing feature from patternfly and it should be implemented on their side. I see no point of adding tests to something that should not even exist in our codebase...

karelhala:
@martinpovolny this is kinda hard to test, because Karma browser handels rgba differently and treats `rgba(${r}, ${g}, ${b}, 0.5)` as floating point and transfers `0.5` into `0.48..` which is not correct. There are unit tests which are testing the color filter, but sadly no integration tests for correct color based on given data.

skateman:
@karelhala everyone loves phantomjs :wink:

PR #288, opened 2018-04-25
Send ascending true or false, if undefined set to false for report data

karelhala:
### Fixes wrong order when switching pages
When switching pages and order of sort is set to DESC it's forgotten and used ASC. This PR fixes such issue by sending `is_ascending` in params, undefined value is mapped to false.

### BZ
https://bugzilla.redhat.com/show_bug.cgi?id=1481840


himdel:
(Restarting travis, the failure is the one solved in https://github.com/ManageIQ/ui-components/pull/290.)

LGTM

PR #291, opened 2018-04-26
Use the name of the category that comes from the API as category_name

eclarizio:
Certain categories have different names than their description, and so sometimes it was failing when trying to choose them to use. This fixes the issue with [this PR from the classic UI](https://github.com/ManageIQ/manageiq-ui-classic/pull/3865) so that we get the correct name from the API in the first place.

The Classic UI PR needs to be merged first.

https://bugzilla.redhat.com/show_bug.cgi?id=1570613

/cc @gmcculloug @d-m-u 

@miq-bot assign @himdel  

### DIFF
--- a/src/dialog-editor/components/modal/modalComponent.ts
+++ b/src/dialog-editor/components/modal/modalComponent.ts
@@ -249,7 +249,7 @@ class ModalController {
     _.forEach(this.categories.resources, function (name) {
       if(name['id'] === item) {
         vm.modalData.options.category_description = name['description'];
-        vm.modalData.options.category_name = _.snakeCase(name['description']);
+        vm.modalData.options.category_name = name['name'];
       }
     });
   }
--- a/src/dialog-editor/components/modal/modalComponent.spec.ts
+++ b/src/dialog-editor/components/modal/modalComponent.spec.ts
@@ -16,9 +16,10 @@ describe('modalComponentSpec', () => {
 
     describe('#setupCategoryOptions', () => {
       it('sets the id, name and description entries for the selected tag control', () => {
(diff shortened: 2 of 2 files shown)


h-kataria:
change looks good to me.

gmcculloug:
@dclarizio Can you review and merge if good?

src/dialog-editor/components/modal/modalComponent.ts
```diff
@@ -249,7 +249,7 @@ class ModalController {
     _.forEach(this.categories.resources, function (name) {
       if(name['id'] === item) {
         vm.modalData.options.category_description = name['description'];
-        vm.modalData.options.category_name = _.snakeCase(name['description']);
+        vm.modalData.options.category_name = name['name'];
```
syncrou: @eclarizio - There had to be a reason we decided to put `_.snakeCase` in there before right?  https://github.com/ManageIQ/ui-components/commit/8d68195c959da670a513baf5386a57b4aa91ceb3#diff-50f2063e2c8663d66ad3f5dd8131f2fdR21

I thought snakeCase was needed for a lookup on the backend ( at least when that PR was built. )

syncrou: Nevermind - I see that id is coming in from the ui-classic PR you referenced.

gmcculloug: @syncrou Yes, we do use name on the backend for a lookup, but name and description are not required to be the same.  For example the name for `Cost Center` is `cc` and the name for `Workload` is `function`.  So using snakeCase here returns a name that fails the backend lookup.

PR #279, opened 2018-04-17
Fix for default value containing all values of a list instead of single value

eclarizio:
The change is very simple but I broke it out into a function refactor as well since it becomes much easier to test this way.

The effective change is:
```
-    this.dialogValues[field] = data.values;
+    this.dialogValues[field] = data.default_value;
```
In the code below, it's line 207 that gets changed for line 222

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1568440

@miq-bot add_label gaprindashvili/yes, bug
@miq-bot assign @himdel 

@chalettu Can you review please?

PR #292, opened 2018-04-26
Fix for default value being set incorrectly for non 'DialogFieldSortedItem' field types

eclarizio:
Ok, so it turns out the fix for https://github.com/ManageIQ/ui-components/pull/279 was a bug only for sorted items, because sorted items use a default value to show what is selected. However, for something like a textbox/checkbox/date element, "values" are used. So, this will use default_value for sorted items but values for non-sorted items.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1568440

@chalettu One more review, please? 😄 

@miq-bot assign @himdel 

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogUser.ts
+++ b/src/dialog-user/components/dialog-user/dialogUser.ts
@@ -219,7 +219,11 @@ export class DialogUserController extends DialogClass implements IDialogs {
 
   private refreshFieldCallback(field, data) {
     this.dialogFields[field] = this.updateDialogFieldData(field, data);
-    this.dialogValues[field] = data.default_value;
+    if (this.isASortedItemDialogField(data.type)) {
+      this.dialogValues[field] = data.default_value;
+    } else {
+      this.dialogValues[field] = data.values;
+    }
     this.dialogFields[field].fieldBeingRefreshed = false;
 
     this.saveDialogData();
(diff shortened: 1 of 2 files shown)


src/dialog-user/components/dialog-user/dialogUser.ts
```diff
@@ -233,6 +237,18 @@ export class DialogUserController extends DialogClass implements IDialogs {
...
+  /**
+   * Determines if the given field type is a subclass of DialogFieldSortedItem
+   * @memberof DialogUserController
+   * @function isASortedItemDialogField
+   * @param fieldType {string} This is the field type that should be used for comparison
+   */
+  private isASortedItemDialogField(fieldType) {
+    return fieldType === 'DialogFieldDropDownList' ||
+           fieldType === 'DialogFieldRadioButton' ||
+           fieldType === 'DialogFieldTagControl';
```
syncrou: @eclarizio - Is this a list that would have to be manually modified if there is a new subclass of `DialogFieldSortedItem`, something that could be missed in the future?

eclarizio: Yes, that's possible. Not exactly sure the best way around that at the moment, unfortunately. Luckily, in the time I've known dialogs, we've never added another type (and it's been 4 years), so, I highly doubt we'll be adding another subclass for sorted items any time soon 😀

syncrou: I can get behind that. 4 years seems appropriate for Subclass additions.

PR #293, opened 2018-05-03
Fix Dialog Editor styling issues

epwinchell:
This PR is a follow-up to https://github.com/ManageIQ/manageiq-ui-classic/pull/3668 and addresses the following requests:

-  correctly applies form-horizontal class for label positioning
-  decrease size of label column 
-  increase size of text area
-  decrease size of dropdowns
-  move refresh button below text area

https://bugzilla.redhat.com/show_bug.cgi?id=1553697

Old (Classic UI)
screen shot 2018-03-27 at 2 22 15 pm

New (Classic UI)
screen shot 2018-03-27 at 2 20 51 pm

Old (Classic UI)
screen shot 2018-05-02 at 12 55 48 pm

New (Classic UI)
screen shot 2018-05-03 at 10 25 26 am

Old (Service UI)
screen shot 2018-04-11 at 1 38 03 pm

New (Service UI)
screen shot 2018-04-11 at 1 07 21 pm

PR #275, opened 2018-03-28
Track by $index ensures no duplicates in select list

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1558271

### proof, cuz people like that 🐰 
pfselect

PR #295, opened 2018-05-10
Use new 'miq-select' directive instead of patternfly

eclarizio:
Tagging @himdel for some assistance on this one.

~~I basically can't seem to get the template to actually convert the `<select>` tags to the way bootstrap does it with the `<ul>` structure. If I'm in the classic UI and I manually call something like `$('select').selectpicker()`, it works, but the directive should be doing that.~~

~~You can even see in the specs, the spec on line 29 that is doing the expect on the `select.text()` isn't the one that fails, it's failing trying to find the length of the "bootstrap" select elements it's supposed to generate, and then failing on all the other specs assessing the other properties of those elements.~~

The issue is kind of a corner case, but the reason for this change is mainly due to the fact that we are using the `data-live-search` feature for dropdowns, which means we need to include `data-token` on the options (which in turn means we cannot use `ng-options`).  Because of this, when the options in the drop down change, but not the values (for example going from `[["1", "one"], ["2", "two"]]` to `[["1", "uno"], ["2", "dos"]]`), the drop down does not trigger a `selectpicker('refresh')` and the user sees incorrect options to choose from until they pick one and then it magically changes, leaving them confused.

This change will allow the user to specify a new option on the select tag called `watch-model` which angular will place a watch on so that when it changes, `selectpicker('refresh')` gets called.

This also has an added benefit that it consolidates the two different patternfly selects into a single one so we don't have to worry about using patternfly v3 on ops UI and v4 on SUI (at least for selects).

https://bugzilla.redhat.com/show_bug.cgi?id=1574668

### DIFF
--- a/src/miq-select/index.ts
+++ b/src/miq-select/index.ts
@@ -1,0 +1,7 @@
+import MiqSelect from './miqSelect';
+import * as angular from 'angular';
+
+module miqSelect {
+  export const app = angular.module('miqStaticAssets.miqSelect', []);
+  app.directive('miqSelect', MiqSelect);
+}
--- a/application-settings.js
+++ b/application-settings.js
@@ -17,6 +17,7 @@ module.exports = {
     quadicon: '/quadicon',
     treeView: '/tree-view',
     treeSelector: '/tree-selector',
+    miqSelect: '/miq-select',
   },
   nodePackages: 'node_modules/',
   get stylesheetPath() {
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,6 +1,6 @@
 
 const webpackConfig = require('./webpack.config');
-const fileGlob = 'src/**/*.spec.ts';
+const fileGlob = 'src/**/*.spec.[jt]s';
 const vendor = 'dist/js/vendor.js';
 const applicationFile = 'dist/js/ui-components.js';
 const jsonGlob = {pattern: 'src/**/*.json', watched: true, served: true, included: false};
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -50,10 +50,11 @@
 
     <div class="col-sm-4" ng-switch-when="DialogFieldDropDownList">
       <!-- Dropdown field where a single value is expected - PF 3 compatible-->
-      <select pf-select
+      <select miq-select
               data-live-search="true"
-              ng-if="!vm.dialogField.options.force_multi_value && vm.patternflyVersion === 3"
+              ng-if="!vm.dialogField.options.force_multi_value"
               ng-model="vm.dialogField.default_value"
+              watch-model="vm.dialogField.values"
               ng-blur="vm.validateField()"
               ng-change="vm.changesHappened()"
               ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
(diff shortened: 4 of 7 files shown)


src/miq-select/index.ts
```diff
@@ -0,0 +1,7 @@
+import MiqSelect from './miqSelect';
+import * as angular from 'angular';
+
+module miqSelect {
+  export const app = angular.module('miqStaticAssets.miqSelect', []);
+  app.directive('miqSelect', MiqSelect);
```
himdel: `MiqSelect` is `undefined` here

himdel:
@eclarizio Looks like the component works just fine, all you have to do is actually include the right module...

In ui-classic:

```
diff --git a/app/assets/javascripts/miq_angular_application.js b/app/assets/javascripts/miq_angular_application.js
index 5fdc79960..9765c6d39 100644
--- a/app/assets/javascripts/miq_angular_application.js
+++ b/app/assets/javascripts/miq_angular_application.js
@@ -8,12 +8,13 @@ ManageIQ.angular.app = angular.module('ManageIQ', [
   'miqStaticAssets.dialogEditor',
   'miqStaticAssets.dialogUser',
+  'miqStaticAssets.miqSelect',
   'miqStaticAssets.treeSelector',
   'miqStaticAssets.treeView',
diff --git a/app/views/dashboard/show.html.haml b/app/views/dashboard/show.html.haml
index a07581316..e65a067ce 100644
--- a/app/views/dashboard/show.html.haml
+++ b/app/views/dashboard/show.html.haml
@@ -19,3 +19,14 @@
 - if WidgetPresenter.chart_data.present?
   :javascript
     ManageIQ.charts.chartData = #{{"widgetchart" => WidgetPresenter.chart_data}.to_json.html_safe};
+
+
+#magic
+  %select{'miq-select' => '',
+          'ng-model'   => 'foo'}
+    %option{:value => nil}
+      Foo
+    %option{:value => 123}
+      Bar
+:javascript
+  miq_bootstrap('#magic')
```

Works just fine :)

(Still looking into the test failure though.)

himdel:
Tests seem to pass with 

```diff
diff --git a/src/miq-select/miqSelect.spec.js b/src/miq-select/miqSelect.spec.js
index 561022c..00f4253 100644
--- a/src/miq-select/miqSelect.spec.js
+++ b/src/miq-select/miqSelect.spec.js
@@ -2,7 +2,7 @@ describe('miq-select', function () {
   var $scope;
   var $compile;
 
-  // beforeEach(module('miqStaticAssets.miqSelect'));
+  beforeEach(angular.mock.module('miqStaticAssets.miqSelect'));
 
   beforeEach(inject(function (_$rootScope_, _$compile_) {
     $scope = _$rootScope_;
```

eclarizio:
@himdel I think this is ready. 
@chalettu Mind giving this a quick review and possibly sanity checking it for me in the SUI? It appears to be working for me but getting rid of the patternfly version differences I'm worried there's something I may have overlooked.

@miq-bot remove_label wip

chalettu:
I will take a look and load it up a little later today.  The patternfly version differences exist because service UI uses PF4 and Classic UI uses PF3 and the select component in PF was a breaking change between versions.   Removing it in favor of some other select component should be fine but I will validate and let you know.

himdel:
@chalettu Any success testing in SUI?

@eclarizio The conflict comes from https://github.com/ManageIQ/ui-components/pull/293, you may need to rebase..

eclarizio:
@miq-bot remove_label unmergeable

himdel:
Backported to gaprindashvili:

```
commit 9e4e77f575549fe57d3a85765f077235c6cfb3b0 (HEAD -> gaprindashvili)
Date:   Tue May 22 16:23:53 2018 +0200

    Merge pull request #295 from eclarizio/BZ1574668
    
    Use new 'miq-select' directive instead of patternfly
    
    (cherry picked from commit 0baac157df87a393b7bb024f237eec0d21bdb79f)
```

himdel:
This reintroduces the bug that https://github.com/ManageIQ/ui-components/pull/275 was fixing... fixed again in ....(TODO)

PR #310, opened 2018-06-26
Added a functionality to sort dropdown values manualy

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1594301

screencast from 2018-06-26 16-36-28

PR #311, opened 2018-06-27
[GAPRINDASHVILI] Corrected description / values for dropdown list

romanblanco:
Corrections after changes in https://github.com/ManageIQ/ui-components/pull/310

- switched labels for description / value
- update select box after changing the values order manually

screencast from 2018-06-27 13-16-31

This change goes to `gaprindashvili` branch.
I'll add a new PR with the changes done in #310 together with some [specs](https://github.com/ManageIQ/ui-components/pull/310#issuecomment-400359683) to `master`.

@himdel @ZitaNemeckova

### DIFF
--- a/src/dialog-editor/components/modal-field-template/drop-down-list.html
+++ b/src/dialog-editor/components/modal-field-template/drop-down-list.html
@@ -68,14 +68,14 @@
           <input id="description-{{$index}}"
                  ng-blur="vm.entriesChange()"
                  name="description"
-                 ng-model="entry[1]"
+                 ng-model="entry[vm.DROPDOWN_ENTRY_DESCRIPTION]"
                  type="text"
                  placeholder="Description"
                  required/>
           <input id="value-{{$index}}"
                  ng-blur="vm.entriesChange()"
                  name="value"
-                 ng-model="entry[0]"
+                 ng-model="entry[vm.DROPDOWN_ENTRY_VALUE]"
                  type="text"
                  placeholder="Value"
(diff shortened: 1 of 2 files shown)


himdel:
@chalettu Please review.

These indexes have changed back and forth so many times I have no confidence in this.

eclarizio:
I agree with @chalettu, I'm not exactly sure what this PR is attempting to solve with the switching of the description/value stuff. If it's the label that needs switching, we should probably make a new html element instead of changing `ng-model`.

I do get the 2nd part of the change, in that the `default_value` dropdown should react accordingly to changes in the order of the values. 👍 

As @himdel mentioned, #310 for sorting fields manually should be on master as well as Gaprindashvili but I'm not sure the protocol at this point since #310 has already been merged.

romanblanco:
> As @himdel mentioned, #310 for sorting fields manually should be on master as well as Gaprindashvili but I'm not sure the protocol at this point since #310 has already been merged.

@eclarizio I'll cherry-pick the commits to a new PR that will go to `master` branch.

chalettu:
Ok, @romanblanco .   I think the only advice I would give for this PR is to create constants that represent the array indexes so it makes the code a bit more readable.  An outsider will have no idea what is in array index 0 or index 1.  Other than that, it looks good.

martinpovolny:
@romanblanco : I was trigger happy and merged this one. Then I merged a revert so no harm should be done, but please, create a new PR with these changes. Thx!

PR #315, opened 2018-07-03
[GAPRINDASHVILI] Corrected description / values for dropdown list

romanblanco:
https://github.com/ManageIQ/ui-components/pull/311#issuecomment-402149263

___

Corrections after changes in https://github.com/ManageIQ/ui-components/pull/310

- switched labels for description / value
- update select box after changing the values order manually

screencast from 2018-06-27 13-16-31

This change goes to `gaprindashvili` branch.
I'll add a new PR with the changes done in #310 together with some [specs](https://github.com/ManageIQ/ui-components/pull/310#issuecomment-400359683) to `master`.

@himdel @ZitaNemeckova

### DIFF
--- a/src/dialog-editor/components/modal-field-template/drop-down-list.html
+++ b/src/dialog-editor/components/modal-field-template/drop-down-list.html
@@ -68,14 +68,14 @@
           <input id="description-{{$index}}"
                  ng-blur="vm.entriesChange()"
                  name="description"
-                 ng-model="entry[1]"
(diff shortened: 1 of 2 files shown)


Hyperkid123:
I would like to point out that the data format of response is confusing. Key => value structures should be objects not array inside array. That way we would not have to guess which is which... Someone should maybe fix it.

ZitaNemeckova:
Released in v1.0.33

PR #317, opened 2018-07-11
Use separate 'changesHappened' update call for datetimepicker

eclarizio:
Date & time picker was trying to use one model for both input fields, so when the date would change, the time would basically get reset, and when the time changed, the date would get reset so the user could never actually pick the time and date they wanted to pick.

This PR separates the date field and the time field into separate models but joins them back up when changes are made so that the correct value is sent to the API when the user presses 'submit'.

Ultimately, I think we should probably look into a whole "datetimepicker" solution, such as https://www.npmjs.com/package/angular-ui-bootstrap-datetimepicker.

Interestingly enough, patternfly *does* have a datetimepicker that I found (https://www.patternfly.org/pattern-library/forms-and-controls/date-and-time/), however, I don't think it's angular supported yet as I couldn't find much information about it.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1583177

@miq-bot assign @himdel 

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -160,8 +160,8 @@
         <p class="input-group">
           <input uib-datepicker-popup type="text"
                  class="form-control"
-                 ng-model="vm.dialogField.default_value"
-                 ng-change="vm.changesHappened()"
+                 ng-model="vm.dialogField.dateField"
+                 ng-change="vm.dateTimeFieldChanged()"
                  is-open="open"
                  datepicker-options="vm.dateOptions"
                  close-text="Close"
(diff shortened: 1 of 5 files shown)


eclarizio:
Maybe, but if that's the case, I'd rather more instances of custom `changesHappened` methods show up first, and then we can refactor after the fact.

JPrause:
@eclarizio @himdel is there anything preventing this from being merged.

eclarizio:
@JPrause I don't think so, unless @chalettu has any other input.

martinpovolny:
@chalettu : can we merge this one?

chalettu:
Should be fine @martinpovolny . Looks good to me.

himdel:
Backported to gaprindashvili:

```
commit f5417672a57874920d4953e868f84d8ec630ea4b (HEAD -> gaprindashvili, upstream/gaprindashvili)
Date:   Tue Jul 17 15:48:26 2018 +0200

    Merge pull request #317 from eclarizio/BZ1583177-TimepickerFix
    
    Use separate 'changesHappened' update call for datetimepicker
    
    (cherry picked from commit 7f24dcae46948d7bdb4fc200975557847cec2719)
```

ZitaNemeckova:
Released in v1.0.33

PR #328, opened 2018-07-20
[GAPRINDASHVILI] Update main & language catalogs for gaprindashvili

mzazrivec:
This one is for gaprindashvili only.


ZitaNemeckova:
Released in v1.0.33

PR #331, opened 2018-07-24
Update gettext catalog and catalog validation

mzazrivec:
* allow angular expressions in gettext strings
* fix i18n in `dialogField.html`
* update gettext catalog

gaprindashvili/no (I'll create a separate PR for gaprindashvili).


himdel:
Corresponding gaprindashvili PR: https://github.com/ManageIQ/ui-components/pull/332

PR #335, opened 2018-07-26
Add gettext:extract to travis build

mzazrivec:
gaprindashvili/yes


himdel:
Pleasse backport together with https://github.com/ManageIQ/ui-components/pull/368, to prevent making travis always green :)

PR #342, opened 2018-09-03
ci/bower-dev .. use -e, reset locale/ first

himdel:
since ManageIQ/ui-components#335, the bower-dev branch didn't work because there was an extra change in locale/ after running the other tests

adding -e so that we can actually see these failures

and cleaning locale/ so that we can checkout the right branch


himdel:
Aand it works :) https://github.com/ManageIQ/ui-components/commits/bower-dev

PR #345, opened 2018-09-04
Convert all pf-select directives to miq-select

skateman:
If we want to move to angular-patternfly v4, this is necessary and according to @himdel the safest solution that can even be backported if needed.

@miq-bot assign @himdel


himdel:
Looks like we need a `'miqStaticAssets.miqSelect'` in `src/dialog-editor/index.ts` too..

PR #354, opened 2018-10-22
Added validation for entries type

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1555170

screencast from 2018-10-22 08-57-53

### DIFF
--- a/src/dialog-editor/services/dialogValidationService.ts
+++ b/src/dialog-editor/services/dialogValidationService.ts
@@ -31,14 +31,21 @@ export default class DialogValidationService {
         field => ({ status: ! _.isEmpty(field.label),
                     errorMessage: __('Dialog field needs to have a label') }),
         field => ({ status: ! ((field.type === 'DialogFieldDropDownList' ||
-                              field.type === 'DialogFieldRadioButton')
-                             && (!field.dynamic && _.isEmpty(field.values))),
+                                field.type === 'DialogFieldRadioButton')
+                               && (!field.dynamic && _.isEmpty(field.values))),
                     errorMessage: __('Dropdown needs to have entries') }),
         field => ({ status: ! (field.type === 'DialogFieldTagControl'
                                && field.category_id === ''),
                     errorMessage: __('Category needs to be set for TagControl field') }),
         field => ({ status: ! (field.dynamic && _.isEmpty(field.resource_action.ae_class)),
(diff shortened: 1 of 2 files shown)


src/dialog-editor/services/dialogValidationService.ts
```diff
@@ -31,14 +31,21 @@ export default class DialogValidationService {
...
                     errorMessage: __('Category needs to be set for TagControl field') }),
         field => ({ status: ! (field.dynamic && _.isEmpty(field.resource_action.ae_class)),
                     errorMessage: __('Entry Point needs to be set for Dynamic elements') }),
+        field => ({ status: ! ((field.type === 'DialogFieldDropDownList' ||
+                                field.type === 'DialogFieldRadioButton')
+                               && (field.data_type === 'integer')
+                               && (!_.chain(field.values)
+                                   .map(dialog_entries => _.toNumber(dialog_entries[0]))
+                                   .every(value => !_.isNaN(value)).value())),
+                    errorMessage: __('Value type is set as Integer, but entries values are not numbers')}),
```
martinpovolny: The text does not look correct. Can some native speaker review? @h-kataria, @eclarizio ?

eclarizio: Yeah, "entries values" is a bit confusing. Here's my suggestion, though it could probably be improved. I'm not sure if we have any sort of standard wording for these kinds of things.

`Value type is set as Integer, but the value entered is not a number` or
`Value type is set as Integer, but the value entered is not an integer`.

martinpovolny:
Hammer?

simaishi:
Hammer backport details:

```
$ git log -1
commit 24c6888b3e3a6b8449f65dfde68fe8d9dec3e2cc
Date:   Fri Nov 9 11:17:52 2018 +0100

    Merge pull request #354 from romanblanco/bz1555170
    
    Added validation for entries type
    
    (cherry picked from commit 3dc84f09895df5c16ca6924a496130b8ca2b40ce)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1555170
```

himdel:
This is causing `TypeError: _.toNumber is not a function` when angular-patternfly brings in lodash 3.10, instead of the 4.17 we want to use - fixed in https://github.com/ManageIQ/manageiq-ui-classic/pull/6449

PR #357, opened 2018-10-31
Disable toolbar dropdown if no item in the group is enabled.

martinpovolny:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1633727

### DIFF
--- a/src/toolbar/components/toolbar-menu/toolbarListComponent.ts
+++ b/src/toolbar/components/toolbar-menu/toolbarListComponent.ts
@@ -52,6 +52,12 @@ export class ToolbarListController implements IToolbarListBindings {
       this.toolbarList.items &&
       this.toolbarList.items.filter((item: IToolbarItem) => !item.hidden).length > 0;
   }
+
+  private isToolbarEnabled(): boolean {
+    return this.toolbarList &&
+      this.toolbarList.items &&
+      this.toolbarList.items.filter((item: IToolbarItem) => item.enabled).length > 0;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


src/toolbar/components/toolbar-menu/toolbarListComponent.ts
```diff
@@ -52,6 +52,12 @@ export class ToolbarListController implements IToolbarListBindings {
       this.toolbarList.items &&
       this.toolbarList.items.filter((item: IToolbarItem) => !item.hidden).length > 0;
   }
+
+  private isToolbarEnabled(): boolean {
+    return this.toolbarList &&
```
himdel: maybe add ` && this.toolbarList.enabled` to move the whole condition here?

JPrause:
@miq-bot add_label blocker

src/toolbar/components/toolbar-menu/toolbarListComponent.ts
karelhala: Yup, that sounds better, plus use `some` so we don't have to iterate over the whole array.
```TS
private isToolbarEnabled(): boolean {
  return this.toolbarList && 
    this.toolbarList.enabled &&
    this.toolbarList.items &&
    this.toolbarList.items.every((item: IToolbarItem) => item.enabled)
}
```

martinpovolny:
@himdel, @karelhala : please, re-review. Thx!

himdel:
LGTM :)

PR #366, opened 2019-02-20
build vendor before yarn run start & test

himdel:
Since yarn run start opens the demo,
and the demo needs the vendor build to have happened,

we may as well run the vendor build during `yarn run start` :).
*And* `yarn run test` because tests also depend on vendor being present.


himdel:
spec is failing because of the error fixed in https://github.com/ManageIQ/ui-components/pull/367

.. except it isn't failing :( 
Error: https://travis-ci.org/ManageIQ/ui-components/builds/495942580#L1536
But green

PR #369, opened 2019-02-25
Dialog Editor - lazy load automate tree

himdel:
Together with https://github.com/ManageIQ/manageiq-ui-classic/pull/5267, this changes how `miq-tree-selector` is used by DialogEditor:

* wraps all the tree options in a `treeOptions` object, so that we don't have to pass around 6 separate vars through the whole component hierarchy

* removes 6 copies of "show the tree, a close button, and the include domains bootstrap switch", moving it into a separate `dialog-editor-tree-selector` component

* previously, the tree was loaded in https://github.com/ManageIQ/manageiq-ui-classic/pull/5267/files#diff-7f0e7be06cf6602a0477d7092bce6ad2L11 before initializing the whole dialog editor

* now, the tree gets loaded when actually showing it, with a loading spinner

* => we're no longer passing treeSelectorData around, but need the `load` function from ui-classic

* take the fqdn from the data, and send it to the `load` function

https://bugzilla.redhat.com/show_bug.cgi?id=1553846

Cc @skateman


himdel:
in input: `Integration/ServiceNow/CMDB/StateMachines/servicenow_create_cmdb`
request: `/tree/automate_entrypoint?fqdn=Integration%2FServiceNow%2FCMDB%2FStateMachines%2Fservicenow_create_cmdb`

himdel:
@miq-bot remove_label wip

Should be ready.. :)

mzazrivec:
1. start creating new service dialog
2. add dropdown
3. click on edit the dropdown
4. switch dynamic to `Yes`, the following error shows in console:
```
angular.js:14961 TypeError: Cannot read property 'includes' of undefined
    at modalComponent.ts:265
    at dialogEditorService.ts:124
    at arrayEach (lodash.js:516)
    at Function.forEach (lodash.js:9344)
    at dialogEditorService.ts:123
    at arrayEach (lodash.js:516)
    at Function.forEach (lodash.js:9344)
    at dialogEditorService.ts:118
    at arrayEach (lodash.js:516)
    at Function.forEach (lodash.js:9344)
```
5. click on `Options` tab
6. click on `Entry point` icon, immediately the whole form gets messed up.

himdel:
@mzazrivec Are you sure you're testing this branch?
(Or rather, doing all of `yarn link ; yarn start` here, and `yarn link @manageiq/ui-components ; bin/webpack` in ui-classic?)

The only `includes` in `modalComponent.ts` is this one: 
https://github.com/ManageIQ/ui-components/pull/369/files#diff-bd05fa6df1e67d8f350ebc242e5c701dR255

And `if (!undefined || !undefined.includes('foo')) return;` should never call `includes`, right?

---

Also, I can't reproduce it...

nope

skateman:
Got the same issue while testing, both branches checked out, yarn linked and webpack ran before.

himdel:
OK, I just tested this again, adding a `console.log` for the version.

I'm getting the same symptoms, but only when on master.
When on my branch, there is no error.

(the form could still be messed up in step 6, without https://github.com/ManageIQ/manageiq-ui-classic/pull/5267 it will just be an infinispinner with a close button and a switch)

himdel:
So please @mzazrivec , @skateman can you try again, and if it happens, click the backtrace and copy the code it's failing on?

mzazrivec:
@himdel I did the same procedure with linking & webpack, but this time I did `yarn install` in `ui-components` and the error I showed above is gone 🤷

PR #371, opened 2019-03-07
[WIP] Collapse custom buttons in the toolbar if they don't fit on screen

skateman:
When there was more than 3 custom button groups, all custom buttons were collapsed into a kebab menu. I'm changing this behavior to be a little more dynamic: 
* If  the toolbar buttons fit into one row on the screen, all good
* If the kebab menu doesn't exist, create it after the custom button groups
* Move the last custom button group from its original location into the kebab
* Go back to the first step

The whole thing is called on the `$onChange` angular lifecycle hook, i.e. any change in the menu will re-run the hook until there are no further changes. Therefore, after there's enough space in the toolbar or no more buttons to move, the loop will stop. The `$onChange` hook is not called on a screen resize, so this is not a truly responsive solution and I rather don't want to add an `onResize` listener if that's okay.

screenshot from 2019-03-07 10-52-43

@miq-bot add_reviewer @himdel 
@miq-bot add_reviewer @Hyperkid123 
@miq-bot add_reviewer @karelhala 

https://bugzilla.redhat.com/show_bug.cgi?id=1683703

### TODO:
* I'm open for ideas how to test this :scream: 
* The styling is a little off, the 2nd level dropdown should be expanded to the left, ping @epwinchell

### DIFF
--- a/demo/data/toolbar.json
+++ b/demo/data/toolbar.json
@@ -127,6 +127,38 @@
       "enabled": true,
       "title": "Power",
       "text": "Power"
+    },
+    {
+      "id": "sep_2",
+      "type": "separator"
+    },
+    {
+      "id": "kebab_example",
+      "type": "kebab",
+      "items": [
+        {
+          "child_id": "middleware_server_reload",
+          "id": "middleware_server_power_choice__middleware_server_reload",
+          "type": "button",
+          "img": "middleware_server_reload.png",
+          "imgdis": "middleware_server_reload.png",
+          "icon": "pficon pficon-restart fa-lg",
+          "name": "middleware_server_power_choice__middleware_server_reload",
+          "hidden": false,
+          "pressed": null,
+          "onwhen": "1+",
+          "data": null,
+          "enabled": true,
+          "title": "Reload these Middleware Servers",
+          "text": "Reload Server",
+          "confirm": "Do you want to reload selected servers?",
+          "url_parms": "main_div"
+        }
+      ]
+    },
+    {
+      "id": "sep_2",
+      "type": "separator"
     }
   ],
   [
--- a/src/toolbar/components/toolbar-menu/toolbar-kebab.html
+++ b/src/toolbar/components/toolbar-menu/toolbar-kebab.html
@@ -2,7 +2,7 @@
   <button uib-dropdown-toggle class="btn btn-link" type="button">
     <span class="fa fa-ellipsis-v"></span>
   </button>
-  <ul uib-dropdown-menu aria-labelledby="dropdownKebab">
+  <ul uib-dropdown-menu aria-labelledby="dropdownKebab" class="dropdown-menu-right">
     <li ng-repeat="kebabItem in vm.kebabItem.items"
         role="{{kebabItem.isSeparator ? 'separator' : 'menuitem'}}"
         ng-class="{
--- a/src/toolbar/components/toolbar-menu/toolbarComponent.ts
+++ b/src/toolbar/components/toolbar-menu/toolbarComponent.ts
@@ -10,6 +10,8 @@ const CUSTOM_ID = 'custom_';
  * @name ToolbarController
  * @param $window {ng.IWindowService} window service for redirecting to non angular pages.
  * @param $location {ng.ILocationService} location service to handle redirect to different angular pages.
+ * @param $timeout {ng.ITimeoutService} timeout service.
+ * @param $element {ng.IRootElementService} access to the element.
  * @param $sce {ng.ISCEService} service for enabling html strings to be html objects injected to page as html and not as
  * string.
  */
@@ -20,6 +22,8 @@ export class ToolbarController {
   /*@ngInject*/
   constructor(private $window: ng.IWindowService,
               private $location: ng.ILocationService,
+              private $timeout: ng.ITimeoutService,
+              private $element: ng.IRootElementService,
               private $sce: ng.ISCEService) {
   }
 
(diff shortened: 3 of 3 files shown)


terezanovotna:
Looks good! This is the behavior we want! Great job!!

epwinchell:
@miq-bot add_reviewer epwinchell

skateman:
This approach isn't really working, as the `onSelect` isn't triggered more than once if the data is being changed. I cannot even trigger it manually, as angular is preventing the rendering to enter into an endless loop.

I have an alternative solution, but it requires some [additional changes](https://github.com/ManageIQ/ui-components/pull/371) in the toolbar builder.

JPrause:
@skateman if this will be able to be backported, can you add the hammer/yes label.

skateman:
@himdel @karelhala do you think it would be safe to backport something like this? I'm not sure.

himdel:
There haven't been that many changes in the toolbar code, so this looks relatively simple to backport.
(assuming this ends up getting used in the first place)

skateman:
still waiting on @martinpovolny with https://github.com/ManageIQ/manageiq-ui-classic/issues/5325

martinpovolny:
> still waiting on @martinpovolny with ManageIQ/manageiq-ui-classic#5325

Sorry it took me so long.

But there's really nothing to wait for as I wrote in the comment in ManageIQ/manageiq-ui-classic#5325

Javascript side is generating the markup. So make the javascript generate `data-custom` if id =~ /^custom_/` and then you will have your HTML attribute.

There's no point in adding anything to the JSON.

skateman:
@martinpovolny this worked:
```js
`this.$element.find('[id^=custom_]').first().parents('div.miq-toolbar-group')`
```
However, the rendering of the icons is not deterministic, most of the time not all icons get rendered in the moment my `toolbarTooHigh()` function gets executed. Therefore, the function might say the toolbar fits into a single row, but after the icons appear it breaks.

I tried to add an extra timeout around my timeout, so I can watch my function timeout while it timeouts, but it did not really help. As far as I understand, this is how angular works and I cannot really get to the point when all the toolbar items are having their final width.

@himdel @Hyperkid123 @karelhala please help me :fire:

himdel:
Well, `$onChanges` has nothing to do with what the browser is doing, as it can even happen before we show the toolbar...

Maybe something like `showOrHide` in the toolbar (js) controller in ui-classic could work?

himdel:
That, or something like using the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API), or dynamically creating CSS to achieve the same.

skateman:
Is it possible to catch from classic js the final layout of the angular component? If yes, we'd have to rx the info back to the component the info. But it is still crazy ...

There's no way to solve this via CSS and the observer API's compatibility with IE is questionable...

himdel:
As far as I'm aware, there is no way to catch reflow events in general.

And as the result of this condition can change on browser resize, when minimizing the menu, when showing/hiding the toolbar,... I don't believe you *can* even catch them all.

I don't see why this is a problem with css, once you know the max width, which you do at run time.

As for IE, the toolbar has to work, but it doesn't have to collapse. Things have to work in IE, but they don't need to work the same way as in current browsers.

skateman:
@himdel adjusting the collapse when resizing the window is like 1% of the case I'd like to cover, IMO it's totally unnecessary. The thing I'm trying to achieve here is to have the minimum amount of custom buttons in a kebab to have the toolbar in one line.

I don't see how this can be easily solved with CSS, the number of items to be in the kebab should be dynamic based on the available space, therefore we can't do the same as we do with the main menu. The only CSS-like solution I see would be to have all menu items both in the kebab and normally and hide some of them based on the available width, but I don't see how CSS could help with that :disappointed:

himdel:
Aah, you're probably right. I keep forgetting that custom buttons don't have constant width, sorry..

I was thinking something along the lines of generating nth-child rules for the first (button width / container width) children.. but ok, not a solution...

So, angular...

* a directive's `link` gets called after the inside DOM was created and attached. But not sure if that wouldn't need a rewrite to not `.show()` and `.hide()` the toolbar after, but use `ng-show=condition` instead.

* `$postLink()` in components should do the same.

skateman:
@himdel `postLink` was the first thing I tried, but it doesn't waits for the child components to be linked, from which we have a lot, therefore it's useless. Also when we do `replace_right_cell`, we keep the component and RX new data into it which doesn't trigger `postLink`..

We could count the inner components inside the group and wait until all are `postLink`-ed but I don't think it would be deterministic and it is a little overkill for solving this issue...

himdel:
OK.. then, I guess either `$interval` (keep adjusting the toolbar every second or so) or Intersection observer (and keep it always collapsed in IE)

himdel:
(Or wait for the React toolbar rewrite, at which point it *mey* be available when React.render ends, and all you'd have to deal with are all the cases where it rerenders :).)

skateman:
Both seem like overkill, IMO the current implementation (collapse all if more than 3 custom buttons) isn't that bad. I don't like the idea of backporting something like this, and this is marked as something to be backported to hammer :confused:

skateman:
@himdel I think @martinpovolny already knows about this issue and he is implementing the new toolbar component accordingly.

Loicavenel:
@skateman can we have an update here?

skateman:
Oh sure, as @himdel wrote above this one can't really be solved in Angular without having a serious performance issues. @martinpovolny is rewriting the toolbars into React where this feature will be already included, but that one can't really be backported.

himdel:
=> https://github.com/ManageIQ/react-ui-components/pull/115

PR #364, opened 2019-02-20
karma - switch backend from phantomjs to jsdom

himdel:
mostly because phantomjs started breaking locally,
but also jsdom is used by jest

PR #374, opened 2019-03-18
Fix linter issues in specs

skateman:
Any better ideas are welcome ...
@karelhala @martinpovolny @Hyperkid123

### DIFF
--- a/src/common/components/sortItemsComponent.spec.ts
+++ b/src/common/components/sortItemsComponent.spec.ts
@@ -2,7 +2,11 @@ import SortItems from './sortItemsComponent';
 import * as angular from 'angular';
 
 describe('Sort items test', () =>  {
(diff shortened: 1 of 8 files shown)


Hyperkid123:
Seems good enough but i have no clue what is this component and what is used for.

himdel:
LGTM too :).

(Though the validation problem can probably be fixed by using `if (this.validation)` instead of `if (_.isObject(this.validation)) {`, because typescript interprets `if (_.isObject(foo))` as narrowing down the type of foo to be `object`.)

The `TypeError: undefined is not an object (evaluating 'this.$newElement[0].parentNode.classList.contains') in dist/js/vendor.js (line 74571)` problem is present even after #364 switching from phantomjs to jsdom, investigating..

himdel:
The failure is caused by using the wrong version of bootstrap-select - https://github.com/ManageIQ/ui-components/pull/375

PR #375, opened 2019-03-18
Fix travis - restrict bootstrap-select to ~1.12

himdel:
Fixes `undefined is not an object (evaluating 'this.$newElement[0].parentNode.classList.contains')` by making sure we don't use too new a version of bootstrap-select. (ui-classic is using `@pf3/select` which is a fork of the 1.12 branch of bootstrap-select)

And just because it still fails with that, this also removes a useless `_.isObject` which narrows the type of `this.validation` from `any` to `object`, causing errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "awesome-typescript-loader": "~4.0.1",
     "bootstrap-combobox": "^1.0.2",
     "bootstrap-datepicker": "^1.6.4",
-    "bootstrap-select": "^1.12.1",
+    "bootstrap-select": "~1.12.1",
     "bootstrap-switch": "^3.3.4",
     "browser-sync": "^2.18.2",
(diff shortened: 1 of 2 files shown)


himdel:
@simaishi the hammer branch is also too optimistic with the dependency, so adding `hammer/yes` as hammer branch will very likely fail the next time travis runs too.

simaishi:
Hammer backport details:

```
$ git log -1
commit 5299c7bd36cedccc7ae6fe089f7fec6525fde66b
Date:   Tue Mar 19 08:52:27 2019 +0100

    Merge pull request #375 from himdel/fix-travis
    
    Fix travis - restrict bootstrap-select to ~1.12
    
    (cherry picked from commit 18535bd7b354a36fd3901256fec5ade39942ae5b)
```

PR #376, opened 2019-03-27
Dialog editor: Make "Load Values on Init" independent of "Show Refresh Button" 

himdel:
Depends on:
  * schema: https://github.com/ManageIQ/manageiq-schema/pull/357 (merged)
  * core: https://github.com/ManageIQ/manageiq/pull/18600 (merged)

Dynamic dialog fields can load on init (once), by manually triggering a refresh button, or from "fields to refresh".

Before, when there was no refresh button, load on init was hardcoded to true on the model side and the switch was hidden.

This makes the "Load Values on Init" switch always visible, independent of "Show Refresh Button".

good

https://bugzilla.redhat.com/show_bug.cgi?id=1684575

### DIFF
--- a/src/dialog-editor/components/modal-field-template/dynamic-values.html
+++ b/src/dialog-editor/components/modal-field-template/dynamic-values.html
@@ -5,7 +5,7 @@
          switch-on-text="{{'Yes'|translate}}"
          switch-off-text="{{'No'|translate }}"/>
 </div>
-<div ng-if="vm.modalData.show_refresh_button" pf-form-group pf-label="{{'Load values on init'|translate}}">
+<div pf-form-group pf-label="{{'Load values on init'|translate}}">
   <input bs-switch
          ng-model="vm.modalData.load_values_on_init"
          type="checkbox"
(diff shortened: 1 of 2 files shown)


himdel:
Cc @gmcculloug, @eclarizio ... an alternative to this would be to allow the user to actually disable both switches (so, no refresh button, and no loading on init), but I'm not sure how useful that would be.

(And would depend on backend changes, because...
```
app/models/dialog_field_sorted_item.rb
159-  def load_values_on_init?
160:    return true unless show_refresh_button
161-    load_values_on_init
162-  end

app/models/dialog_field_text_box.rb
88-  def load_values_on_init?
89:    return true unless show_refresh_button
90-    load_values_on_init
91-  end
```
)

himdel:
EDIT: oh, never mind, turns out this is useful when you only want the field to update from "fields to refresh"

@miq-bot add_label wip

himdel:
@miq-bot remove_label wip

(but dependencies)

himdel:
@miq-bot add_label wip

TODO: make sure the default values still make sense

himdel:
@miq-bot remove_label wip

New fields should get `load_values_on_init` set to true by default.

himdel:
@miq-bot add_label hammer/no

(because https://github.com/ManageIQ/manageiq-schema/pull/357#issuecomment-478056372)

PR #378, opened 2019-04-01
Allow validator_rule to be updated in case of dynamic text boxes

eclarizio:
The UI wasn't changing the validation rule for text boxes when a new rule would come through via automate. This should fix that.

I also updated the message to actually include the regex that the user should match against, otherwise the user is simply guessing at what their "invalid format" is. I'm guessing maybe we had a reason to obscure the regex before, so if that is the case, it can easily be left out of this PR.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1694737

@miq-bot add_label bug, hammer/yes

/cc @tinaafitz Can you comment on the validation message change?

@miq-bot assign @himdel

### DIFF
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -167,7 +167,7 @@ export default class DialogDataService {
         const regex = new RegExp(regexPattern);
         const regexValidates = regex.test(fieldValue);
         validation.isValid = regexValidates;
-        validation.message = __('Entered text does not match required format.');
+        validation.message = __('Entered text should match the format:') + ' ' + regexPattern;
       }
     }
 
--- a/src/dialog-user/components/dialog-user/dialogField.spec.ts
+++ b/src/dialog-user/components/dialog-user/dialogField.spec.ts
@@ -62,7 +62,7 @@ describe('Dialog field test', () => {
       const expectedValue = {
         isValid: false,
         field: 'Service Name',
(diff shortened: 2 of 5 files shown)


tinaafitz:
@eclarizio Thanks, I think the validation message change makes sense.

src/dialog-user/services/dialogData.ts
```diff
@@ -167,7 +167,7 @@ export default class DialogDataService {
         const regex = new RegExp(regexPattern);
         const regexValidates = regex.test(fieldValue);
         validation.isValid = regexValidates;
-        validation.message = __('Entered text does not match required format.');
+        validation.message = __('Entered text should match the format: ' + regexPattern);
```
himdel: `__('Entered text should match the format: ') + regexPattern;`

otherwise, gettext won't find the string

simaishi:
Hammer backport details:

```
$ git log -1
commit fb9f75d35468feb951991a4d3d6424769ed5ea21
Date:   Tue Apr 2 14:45:01 2019 +0000

    Merge pull request #378 from eclarizio/add_validator_rule_attribute_to_update
    
    Allow validator_rule to be updated in case of dynamic text boxes
    
    (cherry picked from commit 851569e401f83b1fc7d8fca2340ed3fefe1ea3c7)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1702064
```

PR #379, opened 2019-04-02
[WIP] DialogUser - support integer keys in dropdowns, refreshing multidropdowns

himdel:
The dialog player component can't currently deal with integer keys, because `<option ng-repeat` will always create strings.
=> switched to `ng-options`

Also, the code path for the refresh button and fields to refresh does not go through the `setupField` logic that converts dropdown default values from json.
=> unifying (TODO) 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693294

@miq-bot add_label wip

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -62,6 +62,7 @@
               class="form-control"
               data-container="body"
               id="{{ vm.dialogField.name }}"
+              miq-options="{ 'data-tokens': vm.fieldTokens }"
               ng-options="value[0] as value[1] for value in vm.dialogField.values">
       </select>
 
@@ -75,6 +76,7 @@
               ng-blur="vm.validateField()"
               ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
               input-id="{{ vm.dialogField.name }}"
+              miq-options="{ 'data-tokens': vm.fieldTokens }"
               ng-options="value[0] as value[1] for value in vm.dialogField.values">
       </select>
     </div>
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -62,7 +62,7 @@
               class="form-control"
               data-container="body"
               id="{{ vm.dialogField.name }}"
-              miq-options="{ 'data-tokens': vm.fieldTokens }"
+              miq-options="{ 'data-tokens': value[0] + ' ' + value[1] }"
               ng-options="value[0] as value[1] for value in vm.dialogField.values">
       </select>
 
@@ -76,7 +76,7 @@
               ng-blur="vm.validateField()"
               ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
               input-id="{{ vm.dialogField.name }}"
-              miq-options="{ 'data-tokens': vm.fieldTokens }"
+              miq-options="{ 'data-tokens': value[0] + ' ' + value[1] }"
               ng-options="value[0] as value[1] for value in vm.dialogField.values">
       </select>
     </div>
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -71,7 +71,7 @@
               ng-if="vm.dialogField.options.force_multi_value"
               ng-model="vm.dialogField.default_value"
               watch-model="vm.dialogField.values"
-              ng-change="vm.changesHappened(item)"
+              ng-change="vm.changesHappened()"
               ng-blur="vm.validateField()"
               ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
               input-id="{{ vm.dialogField.name }}"
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -61,12 +61,8 @@
               ng-disabled="vm.dialogField.read_only || vm.inputDisabled"
               class="form-control"
               data-container="body"
-              id="{{ vm.dialogField.name }}">
-        <option ng-repeat="value in vm.dialogField.values track by $index"
-                data-tokens="{{value[0]}} {{value[1]}}"
-                value="{{value[0]}}">
-          {{value[1]}}
-        </option>
+              id="{{ vm.dialogField.name }}"
+              ng-options="value[0] as value[1] for value in vm.dialogField.values">
       </select>
 
       <select miq-select multiple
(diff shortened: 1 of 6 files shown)


himdel:
@eclarizio What is the value the server *should* receive please?

When submitting a service dialog with a multidropdown, I get a `ServiceTemplateProvisionRequest` with `options: {:dialog=>{"Array::dialog_dropdown_list_1"=>"1,2,3"}` when sending an array of `[1,2,3]`. (Happens after fresh load.)

I also get the same when sending a string `"1,2,3"`. (Happens after updating the value, because there's a `.join` call in JS.)

Aand I get `options: {:dialog=>{"Array::dialog_dropdown_list_1"=>"[1,2,3]"}` when sending an json-encoded array (`"[1,2,3]"`).  (Happens after the refresh button.)

(For now, I'm assuming sending the array is the best option.)

And if we were talking about string values as opposed to integer, would there be a difference? (If I send `["1", "2", "3"]`, the request still ends up looking like `1,2,3`.)

src/dialog-user/components/dialog-user/dialogField.html
```diff
@@ -61,12 +61,8 @@
...
               class="form-control"
               data-container="body"
-              id="{{ vm.dialogField.name }}">
-        <option ng-repeat="value in vm.dialogField.values track by $index"
-                data-tokens="{{value[0]}} {{value[1]}}"
-                value="{{value[0]}}">
-          {{value[1]}}
-        </option>
+              id="{{ vm.dialogField.name }}"
+              ng-options="value[0] as value[1] for value in vm.dialogField.values">
```
eclarizio: Well, this is a bit of a problem because we need `data-tokens`. We had the `ng-options` before, but it does not allow the in-dropdown search function to work without the `data-tokens`, and there's no way that I know of to establish them without doing the `ng-repeat`.

himdel: What do you need it for? There is no code in ui-classic, ui-components, or ui-classic dependencies which would look at data-tokens.

himdel: (Could it be that this was only needed back when angular-patternfly 4 pf-select was used?)

himdel: But if it is still needed, if we can be sure the server doesn't care about types, we could convert everything to string and it could work too.

It's just that I don't see any user of this now.

eclarizio: It's for the dropdown search functionality (`data-live-search="true"`), if a customer has a big huge long list of things in their dropdown, they can type to do a search for things. The problem is that without `data-tokens`, I believe it either searches on the value or the description (can't remember which one), but it needs to be able to search on both, which is why the `data-tokens` attribute needs to exist.

himdel: OK, but, `@pf3/select` does not touch `data-tokens`, and neither does the `miq-select` directive.

So.. I guess what you're saying is that we need to backport the `data-tokens` logic to `@pf3/select`.

himdel: To be clear about this, I don't think `data-tokens` ever worked in ui-classic, but they may have worked in service ui.

eclarizio: @himdel pf is based off of bootstrap as far as I know, and the live-search functionality has been available in dropdowns for a while now. Here's some documentation [from the patternfly side](https://github.com/patternfly/patternfly.github.io/blob/master/components/bootstrap-select/docs/docs/examples.md#live-search) and [from the bootstrap side](https://developer.snapappointments.com/bootstrap-select/examples/#live-search)

himdel: Yeah, sorry about that, I was searching wrong. I can see the code now...

https://github.com/himdel/bootstrap-select/blob/master/js/bootstrap-select.js#L185-L206

So, yeah, this worked, and I'm looking for a way of forcing ng-options to create that..

himdel: ... found one :)

teehee

src/dialog-user/components/dialog-user/dialogField.ts
```diff
@@ -127,6 +121,10 @@ export class DialogFieldController extends DialogFieldClass {
   public refreshSingleField() {
     this.singleRefresh({ field: this.field.name });
   }
+
+  public fieldTokens(element) {
+    return element.innerText + ' ' + element.value.split(/:/)[1];
```
eclarizio: Seems weird to rely on the html, we should have the same thing that `value[0]` and `value[1]` were in `data-tokens` from `this.field.values[0]` and `this.field.values[1]`, right? Or am I misunderstanding what this is doing?

src/dialog-user/components/dialog-user/dialogField.html
eclarizio: Good find 👍

src/dialog-user/components/dialog-user/dialogField.ts
himdel: You're exactly right, I'll try. 
But I haven't found any scope where this could be read from, `angular.element(optionElement).scope().value` gives me undefined.

So we'd have to manually match option elements to entries in the array, to be able to do that, which may be a bit tricky to do right around the empty option, but probably doable :)

also, TODO, `$watchCollection`

himdel: Heheh, so, I got this working with `miq-options="{ 'data-tokens': value[0] + ' ' + value[1] }"` :).
But for now not dealing with the empty option at all, simply..

```js
      var match = attrs.ngOptions.match(/for\s+(.+)\s+in\s+(.+)\s*/);
      if (!match) {
        return;
      }
      var attrName = match[1];
      var attrToWatch = match[2];

      var updateOptions = function (newValue) {
        if (!newValue) {
          return;
        }

        angular.element('option', element).each(function (i, e) {
          var fakeScope = {
            $parent: scope,
            $index: i,
          };
          fakeScope[attrName] = newValue[i];                  // <---

          var miqOptions = $parse(attrs.miqOptions)(fakeScope);
          Object.keys(miqOptions).forEach(function(k) {
            var v = miqOptions[k];
            if (_.isFunction(v)) {
              v = v(e);
            }

            angular.element(e).attr(k, v);
          });
        });
      };

      scope.$watch(attrToWatch, updateOptions);
      scope.$watchCollection(attrToWatch, updateOptions);
```

himdel: Pushed :)

JPrause:
@himdel do you have an update on this PR. We need this for the June 17 build if possible.

himdel:
Yes, working on it, should be ready tomorrow :)

himdel:
Closing, code that didn't go in https://github.com/ManageIQ/ui-components/pull/394 went in https://github.com/ManageIQ/ui-components/pull/399 now.

PR #361, opened 2019-02-11
Update nodejs version that should be used by Travis

ZitaNemeckova:
Update `nodejs` version that should be used by Travis from `6` to `8`.
Error:
```
$ yarn global add typings webpack karma typescript
yarn global v1.3.2
[1/4] Resolving packages...
warning typings@2.1.1: Typings is deprecated in favor of NPM @types -- see README for more information
warning karma > log4js > circular-json@0.5.9: CircularJSON is in maintenance only, flatted is its successor.
[2/4] Fetching packages...
info fsevents@1.2.7: The platform "linux" is incompatible with this module.
info "fsevents@1.2.7" is an optional dependency and failed compatibility check. Excluding it from installation.
error karma@4.0.0: The engine "node" is incompatible with this module. Expected version ">= 8".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
The command "yarn global add typings webpack karma typescript" failed and exited with 1 during .
```

@miq-bot add_label hammer/no, gaprindashvili/no, build

@himdel please review, thanks :)

Note to self: 
If backport is needed try something like this:
```diff
diff --git a/.travis.yml b/.travis.yml
index 08b04a6..92ad57d 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
     - node_modules
 
 before_install:
-  - 'yarn global add typings webpack karma typescript'
+  - 'yarn global add typings webpack karma@3.1.1 typescript'
```

PR #383, opened 2019-04-23
[HAMMER] Force browser-sync to ~ instead of ^, fixes hammer travis

himdel:
to prevent...

    error micromatch@4.0.2: The engine "node" is incompatible with this module. Expected version ">=8".

during yarn. (When using node 6, as travis does.)

This was fixed on master via https://github.com/ManageIQ/ui-components/pull/361, we could do that too, since the builds are using 8.11 now, but this looks like a more proper fix (for hammer), since browser-sync 2.26 is what is requiring that new version of micromatch.


himdel:
Well, ok, it fixes one of the issues on travis.
Still looking into the other one..

himdel:
The remaining travis failure was...

```
Dialog field test #dateTimeFieldChanged calls onUpdate with the correct full date FAILED
	TypeError: undefined is not an object (evaluating 'this.dialogField.options.show_past_dates') (line 9)
```

Fixed on master in https://github.com/ManageIQ/ui-components/pull/367.

I've cherry-picked the commit to make sure this gets green, but if you'd prefer to backport #367 separately, please let me know :)

simaishi:
> I've cherry-picked the commit to make sure this gets green, but if you'd prefer to backport #367 separately, please let me know :)

No problem cherry-picking to this PR. Thanks!

PR #386, opened 2019-05-17
Make sure main points to correct file

ZitaNemeckova:
Make sure `:main` in `package.json` points to correct file so `@manageiq/ui-components` can be moved from asset pipeline to webpack

@miq-bot add_label technical debt, hammer/no, changelog/no

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@manageiq/ui-components",
   "version": "1.2.4",
   "description": "UI components for ManageIQ project",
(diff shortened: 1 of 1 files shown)


ZitaNemeckova:
@miq-bot assign @himdel

himdel:
LGTM, makes sense :)

`index.js` doesnt' exist, but `dist/js/ui-components.js` is indeed the file you want when doing `require('@manageiq/ui-components')` :+1:

PR #387, opened 2019-05-21
Use sessionStorage to avoid loss of unsaved dialog

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1707961

The dialog's content is saved into `sessionStorage` while the editor is getting initialized.
Every change should update the content of the saved dialog.
In the `sessionStorage` the dialog is stored under key `service_dialog-new` or `service_dialog-{id}`.
The dialog is removed from the `sessionStorage` if the dialog is submitted, of user confirms he want's to abandon the changes.
In case the user gets unintentionally redirected, an option to restore the changes will be offered.

This PR contains mostly the calls to update the dialog after the change is done.

Links
----------------

* https://github.com/ManageIQ/manageiq-ui-classic/pull/5612
* https://bugzilla.redhat.com/show_bug.cgi?id=1707961

Steps for Testing/QA
-------------------------------

1. Automate - Automation - Customization
2. Edit or create a new dialog
3. Go to any other page in the menu
4. Get back to the dialog editation

### DIFF
--- a/src/dialog-editor/components/modal/modalComponent.ts
+++ b/src/dialog-editor/components/modal/modalComponent.ts
@@ -187,6 +187,9 @@ class ModalController {
       default:
         break;
     }
+    this.DialogEditor.backupSessionStorage(
+      this.DialogEditor.getDialogId(),
+      this.DialogEditor.data);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


himdel:
LGTM, I'm not sure this covers every action the user could do, but it does save every time a field is added/removed/changed in any way :+1:

PR #388, opened 2019-05-23
Dialog editor - fix Validate switch undecided value

himdel:
1. Create a Dialog with text box validation button
2. Give a validation Regex expression
3. Move the button 'No'

Actual results:
it will be again halfway between Yes and No when the user opens the dialog editor next time.

Expected results:
It should be either Yes/No as per previously saved data

--- 

This is caused by the switch using `ng-true-value="'regex'" ng-false-value="undefined"`, together with angular-bootstrap-switch assuming that `null` and `undefined` always mean "not sure" (https://github.com/frapontillo/angular-bootstrap-switch/issues/138)

Changing to use "false" for no validator type. (And to automatically convert null/undefined to false when editing.)

---

Also moved the 4 copies of the switch + regex input to a separate component, and sorted the requires,

https://bugzilla.redhat.com/show_bug.cgi?id=1705013

### DIFF
--- a/src/dialog-editor/components/index.ts
+++ b/src/dialog-editor/components/index.ts
@@ -1,25 +1,27 @@
-import tabList from './tab-list';
 import box from './box';
+import dialogEditor from './dialog-editor';
 import field from './field';
-import toolbox from './toolbox';
 import modal from './modal';
-import modalTab from './modal-tab';
 import modalBox from './modal-box';
 import modalField from './modal-field';
 import modalFieldTemplate from './modal-field-template';
-import dialogEditor from './dialog-editor';
+import modalTab from './modal-tab';
+import tabList from './tab-list';
+import toolbox from './toolbox';
 import treeSelector from './tree-selector';
+import validation from './validation';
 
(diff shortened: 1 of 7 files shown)


himdel:
(backend doesn't need changing, everything is testing `validator_type` for "regex", not explicit nil

himdel:
Cc @eclarizio - if you could review :)

simaishi:
@himdel can this be `hammer/yes`?

himdel:
@simaishi Sure, added :)

simaishi:
Thanks @himdel 

Hammer backport details:

```
$ git log -1
commit 2721a75ebaffd7a38f4b3ba4c24a668b70d69b5c
Date:   Fri May 31 14:41:32 2019 +0200

    Merge pull request #388 from himdel/validation-fix
    
    Dialog editor - fix Validate switch undecided value
    
    (cherry picked from commit b05dd4e53f959dcc084fae01589806fed8d50144)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1717704
```

PR #249, opened 2018-02-06
BZ#1539862 - Created multiselect dialog field type

chalettu:
@miq-bot add_label bug
@miq-bot add_label gaprindashvili/yes

BZ https://bugzilla.redhat.com/show_bug.cgi?id=1539862

PR #390, opened 2019-05-31
Remove ui-select that is no longer used

ZitaNemeckova:
Removing `ui-select` as mentioned in https://github.com/ManageIQ/manageiq-ui-classic/pull/5607#issuecomment-497364608
> LGTM, except it seems this is dead.
>
>ui-select was added in ui-classic in 8fd8d7b, "to match ui-components". (ui-classic has never used ui-select directly it seems)
>
>ui-componens did use ui-select in dialog-user, until ManageIQ/ui-components@5fcb88d (ManageIQ/ui-components#249), where it was replaced by pf-select (and later miq-select).
>
>Can you change this to remove ui-select please? :) (also from ui-components)

@miq-bot add_label dependencies, hammer/no, changelog/no

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     "tslint-loader": "3.2.0",
     "typescript": "~2.8.3",
     "ui-router": "^1.0.0-alpha0",
-    "ui-select": "0.19.8",
     "url-loader": "^0.5.7",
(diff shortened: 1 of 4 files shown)


ZitaNemeckova:
@himdel please have a look, thanks :)

PR #392, opened 2019-06-10
Mark datepicker field as valid only if it's value is an instance of Date

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1692863
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1702497

Steps to Reproduce from BZ ticket:

1. create a new service dialog with a datepicker field
2. add the service dialog to the catalog
3. in the service catalog select the service dialog and click on order
4. Clear the date from the field and enter "test" in the datepicker field

### DIFF
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -146,7 +146,7 @@ export default class DialogDataService {
           validation.isValid = false;
           validation.message = __('This field is required');
         }
-      } else if (_.isEmpty(fieldValue)) {
+      } else if (_.isEmpty(fieldValue) && !_.isDate(fieldValue)) {
         validation.isValid = false;
         validation.message = __('This field is required');
       }
@@ -161,6 +161,12 @@ export default class DialogDataService {
         validation.isValid = regexValidates;
         validation.message = __('Entered text should match the format:') + ' ' + regexPattern;
       }
+    }
+
+    if (field.type === 'DialogFieldDateControl' &&
+        !(fieldValue instanceof Date)) {
+      validation.isValid = false;
+      validation.message = __('Select a valid date');
     }
 
     if (field.type === 'DialogFieldDateTimeControl') {
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -45,9 +45,9 @@ export default class DialogDataService {
 
     if (field.type === 'DialogFieldDateControl') {
       if (_.isNull(field.default_value) || _.isUndefined(field.default_value)) {
-        field.dateField = field.timeField = new Date();
+        field.dateField = new Date();
       } else {
-        field.dateField = field.timeField = new Date(data.default_value);
+        field.dateField = new Date(data.default_value);
       }
     }
 
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -43,6 +43,14 @@ export default class DialogDataService {
       }
     }
 
+    if (field.type === 'DialogFieldDateControl') {
+      if (_.isNull(field.default_value) || _.isUndefined(field.default_value)) {
+        field.dateField = field.timeField = new Date();
+      } else {
+        field.dateField = field.timeField = new Date(data.default_value);
+      }
+    }
+
     field.default_value = this.setDefaultValue(field);
 
     return field;
@@ -95,12 +103,12 @@ export default class DialogDataService {
       defaultValue = data.values[firstOption][fieldValue];
     }
 
-    if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
-      defaultValue = data.values ? new Date(data.values) : new Date();
-    }
-
     if (data.default_value) {
       defaultValue = data.default_value;
+    }
+
+    if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
+      defaultValue = data.dateField ? new Date(data.dateField) : new Date();
     }
 
     if (data.type === 'DialogFieldDropDownList' && data.options.force_multi_value && data.default_value) {
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -145,6 +145,9 @@ export default class DialogDataService {
     };
     validation.field = field.label;
 
+    let dateControlField = (field.type === 'DialogFieldDateControl');
+    let validDateField = (dateControlField && _.isDate(field.dateField));
+
     if (field.required) {
       if (field.type === 'DialogFieldCheckBox' && fieldValue === 'f') {
         validation.isValid = false;
@@ -154,7 +157,7 @@ export default class DialogDataService {
           validation.isValid = false;
           validation.message = __('This field is required');
         }
-      } else if (_.isEmpty(fieldValue) && !_.isDate(fieldValue)) {
+      } else if (_.isEmpty(fieldValue) && !validDateField) {
         validation.isValid = false;
         validation.message = __('This field is required');
       }
@@ -171,8 +174,7 @@ export default class DialogDataService {
       }
     }
 
-    if (field.type === 'DialogFieldDateControl' &&
-        !(fieldValue instanceof Date)) {
+    if (dateControlField && !validDateField) {
       validation.isValid = false;
       validation.message = __('Select a valid date');
     }
--- a/src/dialog-user/services/dialogData.spec.ts
+++ b/src/dialog-user/services/dialogData.spec.ts
@@ -359,11 +359,12 @@ describe('DialogDataService test', () => {
     });
 
     describe('when the data type is a date control', () => {
-      let dateField = {'type': 'DialogFieldDateControl'};
+      let dateField = {'type': 'DialogFieldDateControl',
+                       'dateField': ''};
 
       describe('when the values are undefined', () => {
         beforeEach(() => {
-          dateField['values'] = undefined;
+          dateField.dateField = undefined;
         });
 
         it('returns a new date', () => {
@@ -379,7 +380,7 @@ describe('DialogDataService test', () => {
 
       describe('when the values exist', () => {
         beforeEach(() => {
-          dateField['values'] = '2017-09-18';
+          dateField.dateField = '2017-09-18';
         });
 
         it('returns a new date based on the values', () => {


JPrause:
@romanblanco can you assign a reviewer. We need this for the 5.10.6 build on Monday, June 17

romanblanco:
@himdel could you do the review?

himdel:
Will do :)

src/dialog-user/services/dialogData.ts
```diff
@@ -171,6 +171,11 @@ export default class DialogDataService {
       }
     }
 
+    if (field.type === 'DialogFieldDateControl') {
+      validation.isValid = (fieldValue instanceof Date);
```
himdel: I'm worried this will fail for dynamic fields - the server can't send a Date object, so we have to convert, and that's called only from `$onInit` it seems, not when updating a field

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1702497 (updated in description)

src/dialog-user/services/dialogData.ts
romanblanco: I believe that `setupField` should take care of converting data to `Date` object, maybe I just need to include `DialogFieldDateTimeControl`, anyway I need to verify this works.

https://github.com/ManageIQ/ui-components/blob/0f871a032a7dfabf244240f99f4f02082f61127a/src/dialog-user/services/dialogData.ts#L38-L44

romanblanco: looks like the Dialog User doesn't work well if the value is updated from Automate.

I've applied a simple Automate method (thanks @d-m-u):

```ruby
new_date = Time.utc(2019, 07, 02)
$evm.object['values'] = new_date
```

I'm not sure if the method is applied correctly, but from what happens I think the problem is in the Dialog User anyway:

1. I cleared out the field so it's empty -> the Submit button correctly gets disabled
2. I checked checkbox that should update the dynamic DialogFieldDateControl field
3. The field is updated with today's date (not `utc(2019, 07, 02)`)  -> the Submit button is still **disabled**

If I pick the date from calendar manually, the button gets **enabled**, as expected.

Now I'm using this to log the data to evm.log:

```ruby
trigger = $evm.root['dialog_trigger']

$evm.log(:info, "======= trigger value #{trigger}")
if trigger == "t"
  new_date = Time.utc(2019, 07, 02)
  $evm.object['values'] = new_date
  $evm.log(:info, "======== new date #{$evm.object['values']}")
else # "f"
  $evm.object['values'] = ''
  $evm.log(:info, "======== clearing up date #{$evm.object['values']}")
end

```
but can't get the log into evm.log nor automate.log, and in `setupField` I'm getting only `undefined`, so I might be using the Automate wrong as well.

romanblanco:
```diff
diff --git a/src/dialog-user/services/dialogData.ts b/src/dialog-user/services/dialogData.ts
index c8fe465..a8a0dd8 100644
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -43,6 +43,14 @@ export default class DialogDataService {
       }
     }

+    if (field.type === 'DialogFieldDateControl') {
+      if (_.isNull(field.default_value) || _.isUndefined(field.default_value)) {
+        field.dateField = field.timeField = new Date();
+      } else {
+        field.dateField = field.timeField = new Date(data.default_value);
+      }
+    }
+
     field.default_value = this.setDefaultValue(field);

     return field;
@@ -95,14 +103,14 @@ export default class DialogDataService {
       defaultValue = data.values[firstOption][fieldValue];
     }

-    if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
-      defaultValue = data.values ? new Date(data.values) : new Date();
-    }
-
     if (data.default_value) {
       defaultValue = data.default_value;
     }

+    if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
+      defaultValue = data.dateField ? new Date(data.dateField) : new Date();
+    }
+
     if (data.type === 'DialogFieldDropDownList' && data.options.force_multi_value && data.default_value) {
       defaultValue = JSON.parse(data.default_value);
     }
```

With these changes I'm getting the values as `Date()` even after using Automate, only the validation doesn't get run.

I'm not receiving any values into `value` attribute, so using `default_value`.

There's a lot of confusion with using `values`, `value` or `default_value`. I'm not sure what exactly is correct, also I think DialogFieldDateTimePicker won't work correctly if it works with `value` instead of `values`.

romanblanco:
update:

the validation works now, the only issue is if the value comes from automate, the form does not refresh by itself.

After hovering the form, the 'Submit' button gets updated:

Screencast from 2019-06-24 12-43-45

If the value is updated manually by the user, `changesHappened` is called. `$doCheck` is called if the new value comes from Automate. I've tried to use 

```javascript
this.onUpdate({ dialogFieldName: this.field.name, value: fieldValue });
```

called by `changesHappened`, expecting the form would get updated, but still no change. I'm still trying to find what's missing.

romanblanco:
It seems that the issue is not introduced by the changes in this PR. Creating a new issue, removing [WIP] label.

src/dialog-user/services/dialogData.ts
eclarizio: Yeah, it is pretty confusing. On the automate side, you want to set the `'value'` key like you're doing above:
`$evm.object['value'] = something`

However, on the ui-components side, for most of the field types, `default_value` is the value that is getting passed back from the refresh API call that we should be looking at to determine what to show to the user after a refresh happens. We changed it in https://github.com/ManageIQ/ui-components/commit/7dcb1f7ca9fc404588cfd8944796bd3ed726c376. For sorted items, `values` is the key we use since it needs to be a list, and `default_value` is simply the one that is selected from that list.

On the ui-components side, because of the way datetime controls work, there's special logic for the date and time parts because the `default_value` comes in as a string (cause it's just a JSON response), and then it gets parsed and separated into a `dateField` and a `timeField` since the controls are separate.

src/dialog-user/services/dialogData.ts
```diff
@@ -43,6 +43,14 @@ export default class DialogDataService {
       }
     }
 
+    if (field.type === 'DialogFieldDateControl') {
+      if (_.isNull(field.default_value) || _.isUndefined(field.default_value)) {
+        field.dateField = field.timeField = new Date();
+      } else {
+        field.dateField = field.timeField = new Date(data.default_value);
```
eclarizio: `timeField` both here and on line 48 I don't think mean anything for a `DialogFieldDateControl`, only if the type is a `DialogFieldDateTimeControl`.

eclarizio:
> If the value is updated manually by the user, changesHappened is called. $doCheck is called if the new value comes from Automate. I've tried to use
> `this.onUpdate({ dialogFieldName: this.field.name, value: fieldValue });`
> called by changesHappened, expecting the form would get updated, but still no change. I'm still trying to find what's missing.


Actually, I think it's similar if you have a text box with an automate method that changes the regex validation it doesn't update until you focus/unfocus the field. So, I would agree with that being a separate issue and limiting this PR to simply not allowing something like `test` to be input into a date field.

martinpovolny:
@eclarizio : please, check again.

src/dialog-user/services/dialogData.ts
```diff
@@ -95,14 +103,14 @@ export default class DialogDataService {
...
-    if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
-      defaultValue = data.values ? new Date(data.values) : new Date();
-    }
-
     if (data.default_value) {
       defaultValue = data.default_value;
     }
 
+    if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
+      defaultValue = data.dateField ? new Date(data.dateField) : new Date();
```
eclarizio: I think my only concern here is if `data.dateField` is going to ever be populated for `DialogFieldDateTimeControl`, because up on line 46, we are only setting `field.dateField` for `DialogFieldDateControl`. Is that a problem?

romanblanco: > because up on line 46, we are only setting field.dateField for DialogFieldDateControl

@eclarizio we're setting the `data.dateField` and `data.timeField` for `DialogFieldDateTimeControl` right above the `DialogFieldDateControl`, on line L38-L44:

```javascript
 38     if (field.type === 'DialogFieldDateTimeControl') {
 39       if (_.isNull(field.values) || _.isUndefined(field.values)) {
 40         field.dateField = field.timeField = new Date();
 41       } else {
 42         field.dateField = field.timeField = new Date(data.values);
 43       }
 44     }
```

himdel: I'm still confused about this change here (moving defaultValue=... *under* the other if).

Previously, `default_value` coming from the server would get used.
Now, only `data.dateField` will ever be used for date/time values.

So... 
a) why not `data.timeField` also?
b) should we really never use `default_value` from the server when dealing with date&time?

romanblanco: > why not data.timeField also?

the `timeField` is only related to `DialogFieldDateTimeControl`, there's no time field in `DialogFieldDateControl`

romanblanco: > should we really never use default_value from the server when dealing with date&time?

we're using the `default_value` on the line 46 that @eclarizio has mentioned

himdel: OK, so `default_value` is now read only in  `setupField` and we're using `dateField` only in `setDefaultValue`.

But, that still doesn't explain why `timeField` gets ignored when in a `if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl')` condition ;).

romanblanco: @himdel you're right. The `(data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl')` condition should be split. I'll update the PR.

himdel: ah, LGTM then, thanks :)

simaishi:
Hammer backport details:

```
$ git log -1
commit 0cdc27e97bdfcbda7aa192213bd6fac3f47e1e22
Date:   Tue Jul 2 11:08:05 2019 +0200

    Merge pull request #392 from romanblanco/bz1692863
    
    Mark datepicker field as valid only if it's value is an instance of Date
    
    (cherry picked from commit ffdd33d7ad457b2f6b895c690716b9bfc1b9bc89)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1702497
    https://bugzilla.redhat.com/show_bug.cgi?id=1726388
```

PR #395, opened 2019-06-24
Store collapsed nodes in tree state and reflect it upon rendering

skateman:
When collapsing a lazily-loaded node in an explorer tree, the state is not being stored in the `sessionStorage`. Therefore, if you do a full page reload, the collapsed node will be expanded again. The solution is to store the collapsed state in the `sessionStorage` as `false` for the given node. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1718457

@miq-bot add_reviewer @himdel 
@miq-bot add_label bug, hammer/yes

### DIFF
--- a/src/tree-view/treeViewComponent.ts
+++ b/src/tree-view/treeViewComponent.ts
@@ -70,7 +70,7 @@ export class TreeViewController {
         preventUnselect: true,
         showBorders:     false,
         onNodeExpanded:  this.storeNodeState(true),
-        onNodeCollapsed: this.storeNodeState(undefined),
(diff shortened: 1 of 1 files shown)


himdel:
This used to be `false` before https://github.com/ManageIQ/ui-components/pull/133/commits/b9c9654fad71666af68fc37a4e0b83cfcd57557e .. looks like that was only a performance optimization, should be safe to undo :+1:

himdel:
LGTM, explorer trees now remember when a node gets collapsed
(Except for cases when the node getting collapsed is active, then it autoexpands again.)

PR #398, opened 2019-06-28
 Corrected values for DialogFieldDateTimeControl

romanblanco:
result of discussion in #392:

https://github.com/ManageIQ/ui-components/pull/392#discussion_r296871830
https://github.com/ManageIQ/ui-components/pull/392#discussion_r298549039

PR #399, opened 2019-07-04
Dialog User - stable validation, consistent refresh and initial load

### LINKED ISSUE #396: Refreshing forms on update from Automate
While updating service dialogs by values from automate, the validation does not get updated automatically. After hover action, the Submit button gets enabled:

Screencast from 2019-06-24 12-43-45

himdel:
This changes how dialog user does field init & validation:

* server updates from inital load *and* from subsequent refreshes go through a `initField` function (which calls `setupField`), allowing for consistent data format transformations on load
  * this requires that setupField be idempotent, adding a check to only transform JSON strings when in string form, and not do that again
 * all validation is done on change, no longer also on blur (the blur validation would not propagate out because any saving is done on change)
* dialog field no longer transforms values - previously we would validate multiselects as an array, then join the array to push to dialog user, then validate the same field, with the joined string instead of an array - any transformation needs to happen when saving
* dateField & timeField are no longer a thing outside dialogField - on change, we save the value to default_value, so it can be validated and pushed outside like everything else; incorporates changes from https://github.com/ManageIQ/ui-components/pull/398
* makes refresh code use the same path, and correctly call `initField` for new values
  * (verified single field refresh goes through initField -> setupField -> DialogField.$onInit)

* (later TODO - maybe `setupField` should not even be used by `DialogField`, always relying on `initField`)
* (later TODO - we should only `initField` once, then `setupField` can just do straighforward server format to client format conversion)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1720245
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1693294
Fixes #396 
Closes #398


himdel:
@eclarizio @romanblanco if you can, please review :) :eyes:

himdel:
@miq-bot remove_label wip

Tests should be green too now

src/dialog-user/components/dialog-user/dialogUser.ts
```diff
@@ -79,19 +83,24 @@ export class DialogUserController extends DialogClass implements IDialogs {
     };
     this.onUpdate({ data: outputData });
   }
+
   public validateFields() {
     const validations = {
       isValid: true,
-      messages: []
+      messages: [], // currently unused, should be?
```
eclarizio: Hmm, maybe we need to standardize on how the HTML expects to use this, then. I could've sworn we had a place in the HTML that showed the overall messages but I guess not. It makes sense (to me, at least) to have individual validation error messages (something like "incorrect regex matching") versus overall validations like "Fields are being refreshed".

I think that's out of the scope of this PR though.

src/dialog-user/services/dialogData.ts
```diff
@@ -108,10 +90,12 @@ export default class DialogDataService {
     }
 
     if (data.type === 'DialogFieldDateControl' || data.type === 'DialogFieldDateTimeControl') {
-      defaultValue = data.dateField ? new Date(data.dateField) : new Date();
+      defaultValue = data.default_value ? new Date(data.default_value) : new Date();
     }
 
-    if (data.type === 'DialogFieldDropDownList' && data.options.force_multi_value && data.default_value) {
+    // FIXME maybe better make sure it's never not string (must come from double call)
```
eclarizio: Maybe putting this conditional into a private method with a more descriptive name would help?

src/dialog-user/components/dialog-user/dialogField.ts
```diff
@@ -50,33 +55,30 @@ export class DialogFieldController extends DialogFieldClass {
    */
   public $doCheck() {
     if (!_.isEqual(this.field, this.clonedDialogField)) {
-      this.clonedDialogField = _.cloneDeep(this.field);
-      if (this.validation) {
-        this.field.fieldValidation = this.validation.isValid;
-        this.field.errorMessage = this.validation.message;
-      }
-      this.dialogField = this.service.setupField(this.field);
+      this.$onInit();
```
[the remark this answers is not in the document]
himdel: makes sense, fixing :)

src/dialog-user/services/dialogData.ts
himdel: I just don't like that we're potentially converting twice, but I think it will need a bit more refactoring to make sure we only run the setup once.

So.. leaving for another PR if you don't mind :)

simaishi:
Hammer backport details:

```
$ git log -1
commit eb52db0dafec141001fe3779bdbdb47a6411c8f0
Date:   Mon Jul 8 14:09:44 2019 -0400

    Merge pull request #399 from himdel/validate-bz1720245
    
    Dialog User - stable validation, consistent refresh and initial load
    
    (cherry picked from commit 55f80b8dd014cb08b239d300ac698de6361cd128)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1727989
    https://bugzilla.redhat.com/show_bug.cgi?id=1727990
```

PR #400, opened 2019-07-04
Allow the selection of a node in DialogEditor + TreeSelector

skateman:
If the domain prefix is included in the automate entry point for a dialog element, it is possible to select the right node in the tree selector. 

@miq-bot add_reviewer @himdel 
cc@pkomanek
@miq-bot add_label bug, hammer/no

https://bugzilla.redhat.com/show_bug.cgi?id=1553846


himdel:
LGTM :)

PR #401, opened 2019-07-11
Dialog user - fix validation for date & date&time when field is required; clearing the fields

himdel:
Fixes some fallout from #399 and some earlier issues:

* date & date&time required fields would now fail validation when filled in, because dates are considered empty `_.isEmpty(new Date()) === true`
* date&time could not get cleared `Cannot read property 'getFullYear' of null`
* date, date&time, and other fields were missing the validation message when failing validation
* cleared date or date&time would fail validation even when not required

Now,
date & date&time should behave the same,
both go to `null` when cleared,
both fail validation when `null` and required,
neither fail validation when non-null or not required.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1728600

Cc @eclarizio , @romanblanco

### DIFF
--- a/src/dialog-user/components/dialog-user/dialogField.html
+++ b/src/dialog-user/components/dialog-user/dialogField.html
@@ -2,12 +2,13 @@
      class="form-group"
      ng-class="{'has-error': vm.validation.isValid === false}">
 
-  <label class=" col-sm-3 control-label">{{ ::vm.dialogField.label }}
+  <label class="col-sm-3 control-label">
+    {{ ::vm.dialogField.label }}
     <i class="fa fa-info-circle primary help-icon"
-         ng-if="vm.dialogField.description" 
-         tooltip-append-to-body="true"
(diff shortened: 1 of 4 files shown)


himdel:
This does change the previous behaviour where an empty date would fail validation even when it was not a required field.

I think that was a bug, but just in case it's not... :)

romanblanco:
:+1:

simaishi:
Hammer backport details:

```
$ git log -1
commit 1d7732078b6fb3ee2b6ef72586529629a723f2a6
Date:   Fri Jul 19 09:22:44 2019 -0400

    Merge pull request #401 from himdel/datetimesubmit-bz1728600
    
    Dialog user - fix validation for date & date&time when field is required; clearing the fields
    
    (cherry picked from commit 28592b743fd4092f1c8b9d4df8e984ba85dc4197)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1731991
```

PR #394, opened 2019-06-17
DialogUser dropdowns - support integer keys

himdel:
(extracted from https://github.com/ManageIQ/ui-components/pull/379)

This adds a `miq-options` directive to allow using `ng-options` and setting additional attributes (`data-tokens`) on the generated `<option>` elements.

And it changes dialog-user dropdowns to use these => numeric values should be supported in dropdowns now.

PR #402, opened 2019-07-11
DialogUser - convert single-select default_value to the right data_type

himdel:
Introduced in https://github.com/ManageIQ/ui-components/pull/394,
we now require `default_value`'s type to match the types of keys in `values`.

But:

* the API never returns a non-string `default_value` - we have to do the conversion to the right type.
  For multiselects, the value is JSON-encoded and preserves types,
  but for single select, even integer values get passed as strings.
* we want to explicitly support converting the keys in `values` to the right type

So now, all of these get converted (on load) to the type in `data_type`:

* `default_value` for single-selects
* each value in `default_value` for multi-selects
* each `value[0]` from `values` for all selects

The conversion rules are as follow:

* if `data_type` is `string`:
  * `null` or `undefined` (actual nil values, not strings) get treated special and stay as `null`
  * any other value gets naively stringified by a simple `toString()`
* if `data_type` is `integer`:
  * `null` or `undefined` (actual nil values, not strings) get treated special and stay as `null`
  * integers stay integers
  * other numbers (floats), or integers over 9007199254740991 may or may not match each other or other such numbers
  * string representations of integers get converted to integers, **including** representations with leading/trailing whitespace (`  1 `), or trailing non-numeric parts (`23 kg`), but **excluding** base-specific notations (`0x123`) or scientific notation (`2e3`)
  * any other value gets converted to 0

A side effect of this change is that `0` gets treated as a valid default dropdown option, so a first option with 0 value will now always get sorted first, whereas previously, this treatment would only apply to non-numeric or null.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1729046

### DIFF
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -17,21 +17,38 @@ export default class DialogDataService {
 
     const sortableFieldTypes = ['DialogFieldDropDownList', 'DialogFieldRadioButton'];
     if (_.includes(sortableFieldTypes, field.type)) {
-      const dropDownValues = [];
-      for (let option of field.values) {
-        const value = ((field.data_type === 'integer' && option[0] !== null) ? parseInt(option[0], 10) : option[0]);
-        const description = (!Number.isInteger(option[1]) ? option[1] : parseInt(option[1], 10));
-        dropDownValues.push([value, description]);
-      }
-      field.values = dropDownValues;
-      if (data.options.sort_by !== 'none') {
-        field.values = this.updateFieldSortOrder(field);
-      }
+      field.values = this.setupSortableValues(field);
     }
 
     field.default_value = this.setDefaultValue(field);
 
     return field;
+  }
+
+  // converts values to the right data_type, and order
+  public setupSortableValues({
+    data_type,
+    default_value,
+    options,
+    values,
+  }) {
+    let dropDownValues = [];
+
+    for (let option of values) {
+      const value = ((data_type === 'integer' && option[0] !== null) ? parseInt(option[0], 10) : option[0]);
+      const description = (!Number.isInteger(option[1]) ? option[1] : parseInt(option[1], 10));
+      dropDownValues.push([value, description]);
+    }
+
+    if (options.sort_by !== 'none') {
+      return this.updateFieldSortOrder({
+        data_type,
+        options,
+        values: dropDownValues,
+      });
+    }
+
+    return dropDownValues;
   }
 
   /**
@@ -42,21 +59,26 @@ export default class DialogDataService {
    * @param data {any} This is a object that is all the information for a particular dialog field
    *
    **/
-  private updateFieldSortOrder(data) {
-    const SORT_DESCRIPTION = 1;
-    const SORT_VALUE = 0;
-    const FIRST_OPTION = 0;
-    const VALUE = 0;
-    const sortBy = (data.options.sort_by === 'value' ? SORT_VALUE : SORT_DESCRIPTION);
-    let tempValues = [...data.values];
+  private updateFieldSortOrder({
+    data_type,
+    options,
+    values,
+  }) {
+    const firstValue = values[0][0];
+    let tempValues = [...values];
     let defaultDropdownField = [];
+
     // The following if deals with a empty default option if it exists
-    if (data.data_type === 'integer' && _.isNaN(tempValues[FIRST_OPTION][VALUE]) ||
-      _.isNull(tempValues[FIRST_OPTION][VALUE])) {
+    if (data_type === 'integer' && _.isNaN(firstValue) || _.isNull(firstValue)) {
       defaultDropdownField = tempValues.shift();
     }
-    let values = _.sortBy(tempValues, sortBy);
-    const sortedValues : any = data.options.sort_order === 'ascending' ? values : values.reverse();
+
+    const sortBy = (options.sort_by === 'value' ? 0 : 1);
+    let sortedValues = _.sortBy(tempValues, sortBy);
+    if (options.sort_order !== 'ascending') {
+      sortedValues = sortedValues.reverse();
+    }
+
     if (defaultDropdownField.length) {
       sortedValues.unshift(defaultDropdownField);
     }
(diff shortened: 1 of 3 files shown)


gmcculloug:
@miq-bot add_reviewer @eclarizio

src/dialog-user/services/dialogData.ts
```diff
@@ -45,21 +56,26 @@ export default class DialogDataService {
...
+  }) {
+    let tempValues = [...values];
+
+    // The following deals with an empty default option if it exists
+    const firstValue = values[0][0];
     let defaultDropdownField = [];
-    // The following if deals with a empty default option if it exists
-    if (data.data_type === 'integer' && _.isNaN(tempValues[FIRST_OPTION][VALUE]) ||
-      _.isNull(tempValues[FIRST_OPTION][VALUE])) {
+    if (! firstValue) {
```
himdel: TODO: test fails here because this also works for actual zeroes now

but, we don't have NaN after the conversion anymore, only zeroes or null
should we convert non-numeric strings to null instead of 0 when converting to integer?

src/dialog-user/services/dialogData.ts
```diff
@@ -93,10 +109,15 @@ export default class DialogDataService {
...
 
-    // FIXME maybe better make sure it's never not string (must come from double call)
-    if (data.type === 'DialogFieldDropDownList' && data.options.force_multi_value
-      && data.default_value && _.isString(data.default_value)) {
-      defaultValue = JSON.parse(data.default_value);
+    // don't convert twice, FIXME: later refactor so that this doesn't get called twice for the same data
+    if (data.type === 'DialogFieldDropDownList' && data.default_value && _.isString(data.default_value)) {
+      if (data.options.force_multi_value) {
+        // multi-select - convert value from JSON, assume right type
+        defaultValue = JSON.parse(data.default_value);
```
himdel: TODO: map these as well

src/dialog-user/services/dialogData.ts
```diff
@@ -17,26 +17,37 @@ export default class DialogDataService {
...
 
+  // converts values to the right data_type, and order
+  public setupSortableValues({
+    data_type,
+    options,
+    values,
+  }) {
+    let dropDownValues = values.map((option) => {
+      const value = this.convertDropdownValue(option[0], data_type);
+      const description = (!Number.isInteger(option[1]) ? option[1] : parseInt(option[1], 10));
```
eclarizio: I don't think we need to convert the description, do we? The `default_value` I believe should always be looking at the `values` and never the description, right? Also, even if we did want to convert the description, could we not just use the new `convertDropdownValue` function?

src/dialog-user/services/dialogData.ts
eclarizio: Hmm, what do you think makes more sense? I think in either case the sort order is just going to be messed up anyway because if you have multiple things that convert into 0 or null, it will probably just keep the order they're already in. I think maybe we convert into 0 simply so that it's "guaranteed" that they get an integer, since that is what they asked for by setting the data_type to integer (even if it wasn't meant to be that).

src/dialog-user/services/dialogData.ts
himdel: This is not new, we were already converting the description - that comes from https://github.com/ManageIQ/ui-components/pull/265.

I *think* this was meant to fix sorting by description when you have all numeric labels, so I just left it as is, without any conversion.

If we use `convertDropdownValue` here, we need to specify that `data_type` also applies to description, not only label, that doesn't sound quite right.

src/dialog-user/services/dialogData.ts
himdel: Well, this whole bit of code is to support options like `{nil => "Nothing selected", 123 => "Default"}`,
and it only looks at the first entry.

So.. the only thing this affects is pretty much that - if you first value is 0, it will end up always getting sorted first, whereas previously only `""`, `null` and `"non-numeric string"` would get treated like that.

himdel: Huh.. and I still have to fix `""` - that should be treated as a `null` also, right?

himdel: @eclarizio WDYT? Should we always treat `""` as `null` or only for integers?

src/dialog-user/services/dialogData.ts
eclarizio: @himdel Ah gotcha, alright, no problem.

src/dialog-user/services/dialogData.ts
eclarizio: @himdel Alright, that makes sense. As far as `""` goes, for an integer I believe on the backend we would be treating it as a "0", it's only if it's like you had above (`{nil => "nothing selected"}`) where it would stay as `nil`. https://github.com/ManageIQ/manageiq/blob/master/app/models/dialog_field_sorted_item.rb#L113

himdel: For the failing spec I guess we have 2 options:

* start treating 0 as a potential default value with integers, fix the spec (it will pass if I replace 0 with 1)
* convert non-numeric strings to null instead of 0

`0` -

eclarizio: I think it's potentially feasible for someone to actually create a hash that has `{0 => "Zero", 1 => "One", 2 => "Two"} #...etc`, so maybe treating 0 as a potential default value with integers and fixing the spec is the way to go. It will also be easier to explain to a customer that "hey, you set the `data_type` to integer so that's why your "foo"s are getting converted to "0"s.

himdel: You're right, that seems to be the safest option :)

himdel:
Ok, @eclarizio if you can review, I think this should be ready, and the description matches the current state.

(I'd still like to write some tests, just to make sure the behaviour is also tested with weird values, not only the strictly valid ones)

himdel:
Thanks :) Fixed a couple more tests which were failing because of the 0 thing.

And maybe I'll leave the other tests for a separate PR so that we can merge & build.

eclarizio:
Yeah that seems fine to me. Nice 👍

simaishi:
Hammer backport details:

```
$ git log -1
commit 2fbdd510fd60cba955839abad2265804a8627f18
Date:   Fri Jul 12 14:32:00 2019 -0400

    Merge pull request #402 from himdel/integer-single-bz1729046
    
    DialogUser - convert single-select default_value to the right data_type
    
    (cherry picked from commit 8b48923eeee45c72ced0d34807f747ce98c1ff2f)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1729594
```

PR #404, opened 2019-07-25
Add a section with creating a new tab

romanblanco:
It does not make sense to create a new tab without a content, with adding a new section, user can drag&drop new Dialog Fields right after creating a new tab.


romanblanco:
@phospi

PR #407, opened 2019-08-01
Dialog editor - validate tag category

himdel:
Adding a Tag control without chosing a category should fail dialog editor validation.

Previously, this didn't work because it's set to `""` not `null` when unset, and it's `field.options.category_id`, not `field.category_id`.

Now, the validation will fail with "'Category needs to be set for TagControl field".

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1729265

---

The problem with that ^ is that we never actually show the validation message anywhere.
We're trying to show it on the submit button in ui-classic, but the button is disabled (and thus not rendering title) when validation fails.

=> Moving validation to the top of the dialog, together with https://github.com/ManageIQ/manageiq-ui-classic/pull/5938 (or https://github.com/ManageIQ/manageiq-ui-classic/pull/5922 for the future)

(merge ui-components first)

### DIFF
--- a/src/dialog-editor/services/dialogValidationService.ts
+++ b/src/dialog-editor/services/dialogValidationService.ts
@@ -1,4 +1,5 @@
 import {__} from '../../common/translateFunction';
+import {sprintf} from 'sprintf-js';
 import * as _ from 'lodash';
 
 const tagHasCategory = (field) => field.options && field.options.category_id;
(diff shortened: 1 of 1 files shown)


himdel:
cc @romanblanco

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 2974d1dde7b8640897b044c4c049589c97ad4043
Date:   Wed Aug 7 13:16:10 2019 +0200

    Merge pull request #407 from himdel/validate-tag-category
    
    Dialog editor - validate tag category
    
    (cherry picked from commit e172cbc3513660e23542bb0c1907a6621b973d00)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1729265
```

PR #417, opened 2019-09-13
fix data table demo: column headers and sorting

bekasovoleg:
**BEFORE**
table_demo

**AFTER**
fixed_table_demo

Also sorting has been fixed


martinpovolny:
ping @himdel : add to Ivanchuk?

himdel:
This is a demo-only change, so, why not :)

martinpovolny:
Ops ;-) I missed the "demo" part :-D

Issue #416, opened 2019-09-13
Broken dropdown with items number in table


bekasovoleg:
Has anyone had such bug? I have this bug when try to use `miq-data-table` in `manageiq-ui-service`
 
![Screenshot from 2019-09-13 18-10-01](https://user-images.githubusercontent.com/14850567/64873542-f7d2e780-d651-11e9-9f35-41eb76748a0b.png)

bekasovoleg:
I figure out there is no `dropdown-menu` class when I use `miq-data-table`
![Screenshot from 2019-09-13 18-18-41](https://user-images.githubusercontent.com/14850567/64874061-fce46680-d652-11e9-8a55-1e20e38a37db.png)

But there is `dropdown-menu` class in `ui-components` Demo

![Screenshot from 2019-09-13 18-15-42](https://user-images.githubusercontent.com/14850567/64873808-8ba4b380-d652-11e9-8709-82f5d47bfc22.png)

himdel:
Fixed by #418, I believe :)

himdel:
> when try to use miq-data-table in manageiq-ui-service

@bekasovoleg side note, we're trying to move off angular (in all the manageiq projects) to reduce the number of frameworks to ideally 1.

So, if you're writing new code, please consider writing it in react now.

As of https://github.com/ManageIQ/manageiq-ui-service/pull/1576, you *can* use react components in ui-service angular screens.

If you need the full GTL, you're out of luck, we don't have a full react implementation *yet*,
but we do have a [SimpleTable](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/components/simple-table/simple-table.jsx) component for just a table,
and a [ReportDataTable](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/components/report-data-table.jsx) component which is a gtl-like table with a paginator.

bekasovoleg:
@himdel it would be great to move `ReportDataTable` to https://github.com/ManageIQ/react-ui-components

martinpovolny:
> @himdel it would be great to move ReportDataTable to https://github.com/ManageIQ/react-ui-components

It talks directly to the `API`. So it depends on manageiq-ui-classic.

Why would you want to move it there?

PR #418, opened 2019-09-13
fix uib-dropdown according to documentation

bekasovoleg:
Fix issue #416

### DIFF
--- a/src/toolbar/components/toolbar-menu/toolbar-list.html
+++ b/src/toolbar/components/toolbar-menu/toolbar-list.html
@@ -7,7 +7,7 @@
     {{vm.toolbarList.text}}
     <span class="caret"></span>
   </button>
-  <ul class="uib-dropdown-menu" role="menu">
+  <ul class="uib-dropdown-menu" uib-dropdown-menu role="menu">
     <li ng-repeat="item in vm.toolbarList.items track by $index" ng-class="{disabled: !item.enabled}">
       <miq-toolbar-click ng-if="item.type !== 'separator'"
                          item="item"


bekasovoleg:
1st dropdown on screenshot has `uib-dropdown-menu` directive and works properly

https://github.com/ManageIQ/ui-components/blob/master/src/common/components/miqPfSort.html#L7


miq_table

himdel:
LGTM, thanks! :)

--- 

Looks like the original code only worked with angular-ui-bootstrap pre-1.0 (class support was removed in https://github.com/angular-ui/bootstrap/pull/4975).

angular-patternfly dropped 0.14 in https://github.com/patternfly/angular-patternfly/commit/ff80d4a41225fbf0e8488887f9c3dc71dc09f18b, but that's v4, we're using v3, which [still depends](https://github.com/patternfly/angular-patternfly/blob/v3.26.0/package.json#L18) on 0.14.

Both ui-classic and ui-service are now using the last angular-ui-bootstrap release - 2.5.6, and have been for about 2 years:

* in ui-classic the version changed in https://github.com/ManageIQ/manageiq-ui-classic/pull/833 from 0.13 to whatever bower installs (presumably still 0.14 then) and got explicitly updated to 2.5.6 during  the switch to npm - https://github.com/ManageIQ/manageiq-ui-classic/pull/4641

* in ui-service, this changed from 0.14 to 2.* in https://github.com/ManageIQ/manageiq-ui-service/pull/363

himdel:
We do have css for `.uib-dropdown-menu` (and `.uib-dropdown-menu-right`), so looks like the class has to stay too :+1:.

Searching for more, ` 'class='\''[^'\'']*uib'` or `'class="[^"]*uib'` in ui-service doesn't find anything,
in ui-components, this finds only:

```
src/toolbar/components/toolbar-menu/toolbar-list.html
2:  <button type="button" uib-dropdown-toggle class="btn uib-dropdown-toggle btn-default"
10:  <ul class="uib-dropdown-menu" uib-dropdown-menu role="menu">

src/site-switcher/site-switcher.html
5:  <div uib-dropdown-menu class="uib-dropdown-menu uib-dropdown-menu-right" aria-labelledby="domain-switcher">
```

And in ui-classic, `\.uib` also finds nothing.

So, looks like found the last of these :+1:

PR #421, opened 2019-09-16
Validate content for non-required fields

romanblanco:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1670051

In case the `DateControl` or `DateTimeControl` field are not set as `required`, the added code tests if the entered value is a date or an empty field and therefore valid.

### DIFF
--- a/src/dialog-user/services/dialogData.ts
+++ b/src/dialog-user/services/dialogData.ts
@@ -204,6 +204,14 @@ export default class DialogDataService {
             }
           }
       }
+    } else {
+      if (field.type === 'DialogFieldDateControl' ||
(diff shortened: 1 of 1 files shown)


src/dialog-user/services/dialogData.ts
```diff
@@ -204,6 +204,14 @@ export default class DialogDataService {
             }
           }
       }
+    } else {
+      if (field.type === 'DialogFieldDateControl' ||
+          field.type === 'DialogFieldDateTimeControl') {
+        if (!(_.isDate(value) || value === null)) {
```
himdel: `null` is valid when not required

PR #429, opened 2019-11-25
[WIP] DialogEditor - angular-ui-bootstrap 2.x compatibility

himdel:
Needed by https://github.com/ManageIQ/manageiq-ui-classic/pull/6449, but can go in standalone.

Since 2.x, angular-ui-bootstrap supports `.component` in addition to `.controller` + `.template` in modals, but that acts very differently (ignores controller and template, and renders the component with a fixed set of args).

The old behaviour still works, but we can't just pass an options object with `.component` to `open`.
Replacing with an object created with just the relevant options. (So, this is pretty much `delete options.component`.)

Also changing a "catch any error and ignore it" to at least `console.error` it,
and because of changes in how ng-switch-when interacts with other directives on the same element, transforming the dialog editor template from (simplified):

```haml
  .col-sm-5
    %input{ng-switch-when...., ng-model...}
    %select{ng-switch-when..., miq-select...}
```

to:

```haml
  .col-sm-5{ng-switch-when....}
    %input{ng-model...}
  .col-sm-5{ng-switch-when....}
    %select{miq-select...}
```

And finally fixing the tag control preview to also support multiselect, and the function used to transform values from JSON in the editor to deal with the values already having been transformed (happened when switching a new dropdown to multiselect).

### DIFF
--- a/src/dialog-editor/components/field/field.html
+++ b/src/dialog-editor/components/field/field.html
@@ -1,36 +1,39 @@
 <div ng-switch on="vm.fieldData.type" class="form-group">
   <label class="control-label col-sm-2">{{ vm.fieldData.label }}</label>
-  <div class="col-sm-5">
 
-    <!-- text box -->
-    <input ng-switch-when="DialogFieldTextBox"
-           ng-model="vm.fieldData.default_value"
+  <!-- text box -->
+  <div class="col-sm-5" ng-switch-when="DialogFieldTextBox">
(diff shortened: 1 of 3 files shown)


himdel:
@miq-bot add_label wip

mostly because there is one more dialog editor select issue,
so possibly this should be split differently

himdel:
Closing, will be part of #6954

PR #434, opened 2020-02-28
Toolbars - clean up, move remnants to gtl

himdel:
UI-classic is using react toolbars since https://github.com/ManageIQ/manageiq-ui-classic/pull/5997.
That means we can remove angular toolbars now.

Except the GTL component is using `miq-toolbar-list` (and thus also `miq-toolbar-click`) for pagination.

So, moving these 2 components over to the gtl/ section, and removing the rest.


himdel:
@miq-bot add_label wip

test in the ui..

himdel:
@miq-bot remove_label wip

The paging toolbar still looks the same, no errors.

PR #440, opened 2020-05-22
dialog-user: add flag for dialog reconfigure

mzazrivec:
In case a dialog is being reconfigured, we need to honor the non-reconfigurable field flag and disable the respective input field.

@himdel 

https://bugzilla.redhat.com/show_bug.cgi?id=1837410


src/dialog-user/components/dialog-user/dialogField.ts
```diff
@@ -49,6 +50,7 @@ export class DialogFieldController {
   public setup() {
     this.clonedDialogField = _.cloneDeep(this.field);
     this.dialogField = this.service.setupField(this.field);
+    this.dialogField.disabled = this.reconfigureMode && !this.dialogField.reconfigurable;
```
himdel, line 53: Can a field be disabled from automate?

I think this bit should only set disabled=true, but not overwrite with false if already present.

mzazrivec: Fixed (as discussed privately).

himdel:
Thanks, LGTM :)

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 90e3d6988512af0bfae0a04670376e2078be07f7
Date:   Tue May 26 12:30:01 2020 +0000

    Merge pull request #440 from mzazrivec/add_support_for_dialog_reconfigure

    dialog-user: add flag for dialog reconfigure
    (cherry picked from commit 5e79adf0217ff1068a5c336e8ef5c291fe2a5933)

    https://bugzilla.redhat.com/show_bug.cgi?id=1837410
```

PR #441, opened 2020-06-15
dialogData#validateField - fail validation when regex is not supported by browser

himdel:
Regex validation in the browser depends on the browser supporting the regex,

when using negative lookbehind, which is not supported in Firefox <78,
it just logs an error in the console:

bad

now, it also shows a message

good-firefox

(when the feautre is supported, no effect)

good-chrome

Related: https://github.com/ManageIQ/manageiq-ui-classic/issues/7050


himdel:
@mzazrivec @simaishi WDYT, can we backport this one to jansa?

jansa ui-components travis is currently failing https://travis-ci.com/github/ManageIQ/ui-components/builds/178817018
and the tslint change would fix it (or I can extract just that change)

PR #445, opened 2020-07-15
Honor sorting parameters in dialog editor

mzazrivec:
This change is for dialog editor to honor sorting options of dropdown & radio button fields:
* sort either by key or by value
* sort values either as strings or as integers

Needs to go in with https://github.com/ManageIQ/manageiq-ui-classic/pull/7202

https://bugzilla.redhat.com/show_bug.cgi?id=1740878

### DIFF
--- a/src/dialog-editor/components/field/field.html
+++ b/src/dialog-editor/components/field/field.html
@@ -65,23 +65,26 @@
     <!-- drop down list -->
     <div ng-switch-when="DialogFieldDropDownList">
       <div ng-if="!vm.fieldData.options.force_multi_value">
-        <select class="form-control" miq-select
+        <select class="form-control"
+                miq-select
+                ng-options="item[0] as item[1] for item in vm.sortedField"
                 ng-model="vm.fieldData.default_value">
           <option value="" translate>None</option>
-          <option ng-repeat="value in vm.fieldData.values" value="{{value[0]}}">{{value[1]}}</option>
         </select>
       </div>
       <div ng-if="vm.fieldData.options.force_multi_value">
-        <select class="form-control" multiple miq-select
+        <select class="form-control"
+                multiple
+                miq-select
+                ng-options="item[0] as item[1] for item in vm.sortedField"
                 ng-model="vm.fieldData.default_value">
-          <option ng-repeat="value in vm.fieldData.values" value="{{value[0]}}">{{value[1]}}</option>
         </select>
       </div>
     </div>
 
     <!-- radio button -->
     <span ng-switch-when="DialogFieldRadioButton">
-      <label ng-repeat="option in vm.fieldData.values"
+      <label ng-repeat="option in vm.sortedField"
              class="radio-inline">
         <input type="radio"
                name="{{vm.fieldData.name}}"
--- a/src/dialog-editor/components/field/fieldComponent.ts
+++ b/src/dialog-editor/components/field/fieldComponent.ts
@@ -9,11 +9,12 @@ import * as angular from 'angular';
  */
 class FieldController {
   public service: any;
+  public sortedField: any;
   public fieldData: any;
   public boxPosition: any;
 
   /*@ngInject*/
-  constructor(private DialogEditor: any) {
+  constructor(private DialogEditor: any, private DialogData: any) {
   }
 
   /**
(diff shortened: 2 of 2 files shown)


src/dialog-editor/components/field/fieldComponent.ts
```diff
@@ -38,6 +38,42 @@ class FieldController {
...
+   * @returns {Array} of sorted field values
+   */
+  public sortedField() {
+    if (this.fieldData.options.sort_by !== 'none') {
+      let tempValues = [...this.fieldData.values];
+      if (this.fieldData.data_type === 'integer') {
+        tempValues.forEach((item, index) => {
+          const val = parseInt(tempValues[index][1], 10);
+          if (!Number.isNaN(val)) {
+            tempValues[index][1] = val;
```
himdel, line 54: should `[1]` be `[sortBy]`?

Otherwise, you're always converting the label, but sometimes not sorting by label.

himdel:
From what I can tell, in dialog-user:

* `setupSortableValues` uses `data_type` to convert `value`, but not `description`
* `description` seems  to assume `Number.isInteger` can return true for strings, so in practice, it gets kept as string, unless an integer is received in the original data (possibly relevant for dynamic fields)
* then `updateFieldSortOrder` just does `sortBy(values, 0 or 1)` + `reverse`

mzazrivec:
@himdel Should be fixed now.

src/dialog-editor/components/field/fieldComponent.ts
```diff
@@ -38,6 +51,16 @@ class FieldController {
     this.DialogEditor.updatePositions(this.getFields(tabId, boxId));
   }
 
+  // /**
```
himdel, line 54: probably not needed anymore?

src/dialog-editor/components/field/field.html
```diff
@@ -65,23 +65,26 @@
...
                 ng-model="vm.fieldData.default_value">
-          <option ng-repeat="value in vm.fieldData.values" value="{{value[0]}}">{{value[1]}}</option>
         </select>
       </div>
     </div>
 
     <!-- radio button -->
     <span ng-switch-when="DialogFieldRadioButton">
-      <label ng-repeat="option in vm.fieldData.values"
+      <label ng-repeat="option in vm.sortedField()"
```
himdel, line 87: looks like there is no such method, should be vm.sortedField without ()?

src/dialog-user/services/dialogData.ts
```diff
@@ -33,9 +33,11 @@ export default class DialogDataService {
     options,
     values,
   }) {
+    // single value selects do contain [null, '<None>']
+    const allIntegers = values.every((item) => (item[0] === null ? true : !isNaN(item[1]) && Number.isInteger(Number(item[1]))));
```
himdel, line 37: `allIntegers` is true when all descriptions except for the null item look like a number

previous behaviour would only treat actual numbers as numbers, is the change intentional?

Sounds like this would leaad to us sorting descriptions numerically whenever data_type is set to integer,
where previously we could only sort descriptions numerically when they were coming from dynamic fields, if ever.

src/dialog-editor/components/field/field.html
mzazrivec: Fixed

src/dialog-editor/components/field/fieldComponent.ts
mzazrivec: Fixed

src/dialog-user/services/dialogData.ts
mzazrivec: I've been back & forth here and in truth, I don't know how the sorting should behave. But since this PR is about fixing dialog editor and friends so that it honors sorting parameters and not about changing sorting behavior, I'm reverting my change to the original.

mzazrivec:
As a side effect, this change also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1741294

h-kataria:
Marked it as an enhancement to match labels in the UI [PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/7202)

simaishi:
Jansa backport details:

```
$ git log -1
commit 73078d9b1a1f1a219cdce02292981dcffb656992
Date:   Mon Aug 24 10:31:16 2020 +0000

    Merge pull request #445 from mzazrivec/honor_sort_parameters_in_dialog_editor

    Honor sorting parameters in dialog editor

    (cherry picked from commit 2c76c33dbc3af59654748c67df0d3f4c242c8158)
```

PR #427, opened 2019-11-12
Use miq-select in dialogUser's tagging fields

skateman:
Making it consistent with the dropdowns and making a step towards fixing https://github.com/ManageIQ/manageiq-ui-classic/issues/6392 and making https://github.com/ManageIQ/manageiq-ui-classic/pull/6400 less buggy.

@miq-bot assign @himdel

PR #446, opened 2020-07-27
[IVANCHUK] Use miq-select in dialogUser's tagging fields

mzazrivec:
This is ivanchuk version of https://github.com/ManageIQ/ui-components/pull/427

We need it in ivanchuk to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1859524


mzazrivec:
> The changes are the same as #427,
> shouldn't we just add `ivanchuk/yes` there?

I tried a cherry-pick locally and it showed conflict.

PR #449, opened 2020-09-03
Fixed table header shift after icon column was removed from list view

h-kataria:
This issue was introduced in https://github.com/ManageIQ/manageiq-ui-classic/pull/7189 Issue existed on the lists with no checkboxes/icons. Table headers were shifted on such screens.

before
Screenshot from 2020-09-03 17-24-34
Screenshot from 2020-09-03 17-23-36

after
Screenshot from 2020-09-03 17-14-02

Screenshot from 2020-09-03 19-43-29

### DIFF
--- a/src/gtl/components/data-table/data-table.html
+++ b/src/gtl/components/data-table/data-table.html
@@ -14,10 +14,8 @@
          ng-if="tableCtrl.rows && tableCtrl.rows.length !== 0">
     <thead>
     <tr>
-      <th class="narrow">
-
-      </th>
-      <th ng-if="$index !== 0"
+      <th ng-if="tableCtrl.countCheckboxes() > 0" class="narrow"></th>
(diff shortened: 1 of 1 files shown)


src/gtl/components/data-table/data-table.html
```diff
@@ -14,9 +14,6 @@
          ng-if="tableCtrl.rows && tableCtrl.rows.length !== 0">
     <thead>
     <tr>
-      <th class="narrow">
-
-      </th>
       <th ng-if="$index !== 0"
```
himdel, line 17: should this `ng-if` also go away? Otherwise we're still skipping the first column name, right?

h-kataria: @himdel fix revised, please re-review

h-kataria: @himdel, not sure how to fix failures on this PR

himdel: @h-kataria this should fix the test failures https://github.com/ManageIQ/ui-components/pull/450

Issue #453, opened 2020-10-27
Truncation Issue in the Dialog Editor with Certain Languages


DavidResende0:
Affects both Japanese and French translations but may also affect others.

<img src="https://user-images.githubusercontent.com/64800041/97309000-e63da680-1837-11eb-8003-b097f0defae4.png" width="700">
<img src="https://user-images.githubusercontent.com/64800041/97309038-edfd4b00-1837-11eb-91e7-d83c17bb1bcf.png" width="100">

PR #454, opened 2020-12-08
Modified CSS to Prevent Truncation Issue in Certain Locales

DavidResende0:
Found in: Automation -> Automate -> Customization -> Service Dialogs

Fixes: https://github.com/ManageIQ/ui-components/issues/453

Preview:

image

### DIFF
--- a/src/styles/dialog-editor-toolbox.scss
+++ b/src/styles/dialog-editor-toolbox.scss
@@ -12,7 +12,7 @@
 
   .static-field-list {
     list-style: none;
-    width: 107px;
+    width: 115px;
     overflow-y: auto;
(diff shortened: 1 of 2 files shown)


himdel:
Are you sure making those wider and making it impossible to wrap text is the best solution?
Specifically, any time a translation gets longer, we'd need to adjust the sizing again.

I think the issue is not that wrapping is allowed, but that the *height* doesn't adjust to the text size when wrapping occurs.

DavidResende0:
@himdel I took your advice and made it so that the height adjusts to the text-wrapping rather than forcing the string on one line. However, I still extended the width of the box slightly to better fit certain German words.

PR #464, opened 2022-07-18
Upgrade node version to 16

jeffibm:
Upgrading package version 
`node` - from `14.0.0` to `16.0.0`
`npm` from `6.0.0 `to` 8.0.0`

Fixes - https://github.com/ManageIQ/manageiq-ui-classic/issues/8146

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 dist: bionic
 language: node_js
 node_js:
-- '14'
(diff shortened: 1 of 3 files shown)


kbrock:
@Fryguy I believe we were holding back on this so we would use the same version of node on all of the miq subprojects.

Now seems like the perfect time to upgrade this to node 16.

Do you see any issues?

PR #466, opened 2022-08-02
Switch to GitHub Actions

Fryguy:
Part of ManageIQ/manageiq#21620

CI: https://github.com/Fryguy/ui-components/actions?query=branch%3Agithub_actions


Fryguy:
@agrare Please review - The test themselves are failing, but without switching to GHA we can't know that to fix it.  This PR gets actions "working" even if it's currently red.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,35 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+  - cron: '0 0 * * *'
```
Fryguy, line 8: Not sure which one we should use here, but I can change in a follow up

agrare (approved):

bin/setup
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+if [ -n "$CI" ]; then
+	yarn global add typings webpack karma typescript
+fi
+
+yarn install
```
Fryguy, line 9: A lot of this will change with yarn 3, but I want to get GHA -> green -> yarn3 -> node16

PR #471, opened 2023-06-28
Introducing embedded_workflow to dynamic select box in service dialogs

jeffibm:
Screenshots in related PR - https://github.com/ManageIQ/manageiq-ui-classic/pull/8844

### DIFF
--- a/src/dialog-editor/components/modal-field-template/drop-down-list.html
+++ b/src/dialog-editor/components/modal-field-template/drop-down-list.html
@@ -114,14 +114,25 @@
 <div ng-if="vm.modalTabIsSet('options') && vm.modalData.dynamic">
   <form class="form-horizontal">
     <div ng-show="!vm.treeOptions.show">
-      <div pf-form-group pf-label="{{'Entry Point'|translate}}">
-        <div class="input-group">
-          <input type="text" class="form-control" ng-value="vm.showFullyQualifiedName(vm.modalData.resource_action)" disabled>
+      <div pf-form-group pf-label="Automation Type" class="entry_point_selector_types">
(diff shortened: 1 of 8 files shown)


Fryguy:
Biggest problem we're going to have here is that I'm not sure the code can be compiled and pushed to npm

Alternatively, we might need to change the core to use ui-components directly via git instead of npm if that's even possible.

src/dialog-editor/components/modal-field-template/drop-down-list.html
```diff
@@ -114,14 +114,25 @@
 <div ng-if="vm.modalTabIsSet('options') && vm.modalData.dynamic">
   <form class="form-horizontal">
     <div ng-show="!vm.treeOptions.show">
-      <div pf-form-group pf-label="{{'Entry Point'|translate}}">
-        <div class="input-group">
-          <input type="text" class="form-control" ng-value="vm.showFullyQualifiedName(vm.modalData.resource_action)" disabled>
+      <div pf-form-group pf-label="Automation Type" class="entry_point_selector_types">
+        <select class="form-control" miq-select ng-model="vm.modalData.automation_type" ng-change="vm.treeOptions.onAutomationTypeChange()">
+          <option value="{{vm.treeOptions.automationTypes.automate}}" translate>Embedded Automate</option>
+          <option value="{{vm.treeOptions.automationTypes.workflow}}" translate>Embedded Workflows</option>
```
Fryguy, line 120: We probably should only show this if the prototype flag is enabled.  This can be found via API.

agrare:
@DavidResende0 or @GilbertCherrie could you review this as well?

DavidResende0 (approved):

PR #486, opened 2024-02-06
Fix for duplicate field names and render validation messages for server

jeffibm:
Before -

https://github.com/ManageIQ/ui-components/assets/87487049/f9b88090-1e73-45d6-94b2-b31f8a2a7ad8

image
image
image

- Edit a service dialog, 
- Click edit field to open the options modal box
- Change a field name to an already used name (duplicate name)
- Click save at the options modal box
- Click save at the service dialog form
- We will get a warning that the field name cannot be used.
- cancel the form.
- select the service dialog again and edit it
- check the field name of the field where you tried to put a duplicate name.
- See that the field name got updated and now we have a duplicate field.

After

https://github.com/ManageIQ/ui-components/assets/87487049/f3b8c739-6b52-45b3-92f2-8278d54fc7d2

image

- The options modal box will check for duplicate field names
- If there are duplicate field names, the save button of the options modal will be disabled.
- Introduced Validation messages in the options modal box. Earlier it was displayed as a title (on mouse hover) of the save button
- For existing records, the service dialog save button will be disabled if there are duplicate field names.
- The options modal's save button will also be disabled unless you change the field name to avoid duplicate.

### DIFF
--- a/src/dialog-editor/components/modal/modalComponent.ts
+++ b/src/dialog-editor/components/modal/modalComponent.ts
@@ -141,23 +141,45 @@ class ModalController {
   }
 
   /**
+   * Check for duplicate field names.
+   * @memberof ModalController
+   * @function modalUnchanged
+   */
+  public fieldDuplicates() {
+    const fields: string[] = [];
+    const {tabId, boxId, fieldId} = this.elementInfo;
+    const dialogTabs = this.DialogEditor.getDialogTabs();
+    this.DialogEditor.getDialogTabs().forEach((tabs, tabIndex) => (
+      tabs.dialog_groups.forEach((group, groupIndex) => (
+        group.dialog_fields.forEach((field, fieldIndex) => {
+          const name =  (tabId === tabIndex && boxId === groupIndex && fieldId === fieldIndex)
+            ? this.modalData.name
+            : field.name;
+          fields.push(name);
+        })
+      ))
+    ));
+    const isDuplicate = fields.filter((item) => item === this.modalData.name).length > 1 ? true : false;
+    this.modalData.validationMessage = isDuplicate ? sprintf(__('%s is used by another field'), this.modalData.name) : undefined;
+    return isDuplicate;
+  }
+
+  /**
    * Check for changes in the modal.
    * @memberof ModalController
    * @function modalUnchanged
    */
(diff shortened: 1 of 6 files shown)


src/dialog-editor/services/dialogValidationService.ts
```diff
@@ -99,12 +100,28 @@ export default class DialogValidationService {
...
+  /**
+   * Run validations across each dialog field elements and check for duplicate field names.
+   * @memberof DialogValidationService
+   * @function validateDialogFields
+   */
+  public validateDialogFields(field: any) {
+    this.dialogFields.push(field.name);
+    const isDuplicate = this.dialogFields.filter((item) => item === field.name).length > 1 ? true : false;
+    if (isDuplicate) {
+      this.invalid.message = `Please change duplicate field name "${field.name}"`;
```
Fryguy, line 112: I think we should present this more directly.  Something like

```suggestion
      this.invalid.message = `Unable to save due to duplicate field name "${field.name}"`;
```

Fryguy: Separate question - is it possible to add i18n support for this line?

Fryguy:
@DavidResende0 Can you also review here?

DavidResende0 (approved):
LGTM, just look into @Fryguy's comments before merging.

src/dialog-editor/components/modal/modalComponent.ts
```diff
@@ -140,24 +140,46 @@ class ModalController {
...
+        group.dialog_fields.forEach((field, fieldIndex) => {
+          const name =  (tabId === tabIndex && boxId === groupIndex && fieldId === fieldIndex)
+            ? this.modalData.name
+            : field.name;
+          fields.push(name);
+        })
+      ))
+    ));
+    const isDuplicate = fields.filter((item) => item === this.modalData.name).length > 1 ? true : false;
+    this.modalData.validationMessage = isDuplicate ? `"${this.modalData.name}" is used by another field` : undefined;
```
DavidResende0, line 163: This should probably be marked for translation.

Fryguy (approved):
LGTM - @DavidResende0 if you are good please merge.  Note that we will have to cut a new release once it's merged.

PR #491, opened 2024-03-08
Disable all fields when a refresh action is in progress

jeffibm:
Note: There is no dependant PR for this.

**Before**
When we select an option from the drop-down list, a refresh action would be executed and we can select another option before the first refresh action is completed.

https://github.com/ManageIQ/ui-components/assets/87487049/4c988e55-5191-456f-afc5-bd234b9f83c1

**After**
When a refresh is in progress, all fields will be disabled.
Even if you try to select an option, it won't work if a refresh action is in progress.
Once the refresh action is complete, then all fields will be enabled.
image

https://github.com/ManageIQ/ui-components/assets/87487049/d535759a-accc-4d72-946e-574a93cb9bcf


Fryguy:
> Note: There is no dependant PR for this.

Well, we'll need to rebuild ui-components and then change the application to pull in the latest one, right?

DavidResende0 (approved):

jeffibm:
> > Note: There is no dependant PR for this.
> 
> Well, we'll need to rebuild ui-components and then change the application to pull in the latest one, right?

Yes, even though we don't have a logic change in ui-classic, we still need to raise a PR with the updated `ui-components ` version in classics package.json.

jeffibm:
Hey @Fryguy , this PR is completed. Can we merge this? how do we build the new ui-components version so that we can raise a new PR in classic?

GilbertCherrie:
This pr is included in the new build: https://github.com/ManageIQ/manageiq-ui-classic/pull/9129

PR #493, opened 2024-03-19
Add a Dockerfile for building the ui-components package

Fryguy:
@agrare Please review.

I verified this all manually and it all works (even all the way up to publishing, but when it did the publish I just did a Ctrl-C when it asked for the OTP.


agrare (approved):
:+1: works for me

PR #517, opened 2025-09-03
Fix drop down default value

GilbertCherrie:
This is a follow up to this pr: https://github.com/ManageIQ/manageiq-ui-classic/pull/9482 to fix the tag control field default values.

These are test fields showing the default values working for the single and multi select dropdown and tag fields.
Screenshot 2025-09-03 at 1 50 23 PM


asirvadAbrahamVarghese:
Reopening to trigger tests

asirvadAbrahamVarghese (approved):
Looks good to me, but the branch is a bit old, @GilbertCherrie can you rebase to kick tests with the latest Yarn(4.14)?

PR #577, opened 2026-04-22
Drop old service badges

Fryguy:
These badges are pointing to services that no longer exist.

[skip ci]

@asirvadAbrahamVarghese Please review.


asirvadAbrahamVarghese (approved):
Lol looks like those URLs are serving casino stuff now

PR #580, opened 2026-04-23
Cleanup useless regular-expression character escape

Fryguy:
@asirvadAbrahamVarghese These are triggering some low sev CodeQL issues because they are unnecessary.


asirvadAbrahamVarghese (approved):
LGTM

PR #584, opened 2026-04-24
Drop line numbers from locale template

Fryguy:
This avoids having to update the locale for little changes that don't actually change the locale strings. We have done this in other parts of the application as well. Note that I would prefer to drop the file names as well, but that's not an option in angular-gettext-tools.

@asirvadAbrahamVarghese Please review.
cc @jrafanie (Do we actually use this or not? I wonder if we should if not)

### DIFF
--- a/locale/ui-components.pot
+++ b/locale/ui-components.pot
@@ -4,648 +4,539 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
-#: ./src/dialog-editor/components/box/box.html:55
+#: ./dist/js/ui-components.js
+msgid "%s is used by another field"
+msgstr ""
+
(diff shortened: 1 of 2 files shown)


asirvadAbrahamVarghese:
Ahh, here we go again

asirvadAbrahamVarghese:
Kicking test

asirvadAbrahamVarghese (approved):
Looks fine to me, @jrafanie do you want to take a look?

jrafanie:
> @asirvadAbrahamVarghese Please review.
> cc @jrafanie (Do we actually use this or not? I wonder if we should if not)

Apparently, we do:

https://github.com/ManageIQ/manageiq/blob/master/lib/tasks/locale.rake#L192-L195

If it's not a Vmdb::Plugin, it has to be in that list.

PR #585, opened 2026-04-24
[WIP] Add ability to unwind AggregateError

Fryguy:
This is what AI did to figure out the AggregateError.

WIP because I want to see if it helps - if we determine it helps I can undo some of these changes to narrow done what is actually required.


Fryguy:
Closing with solution in #586

PR #616, opened 2026-06-01
Add automated locale update workflow

jrafanie:
Creates GitHub Actions workflow to automatically extract translation strings from ui-components and generate locale/ui-components.pot file.

- Runs twice per month one week before ManageIQ's locale updates
- Follows ManageIQ's locale_update_all.yaml pattern
- Creates PRs via miq-bot/ui-components fork

Fixes https://github.com/ManageIQ/manageiq/issues/23856

### DIFF
--- a/.github/workflows/locale_update.yaml
+++ b/.github/workflows/locale_update.yaml
@@ -1,0 +1,55 @@
+name: Locale Update
+on:
+  schedule:
+  - cron: '0 0 8,22 * *'
+  workflow_dispatch:
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+permissions:
+  contents: read
+jobs:
+  locale_update:
+    if: github.repository_owner == 'ManageIQ'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Set up system
+      run: bin/before_install
+    - name: Set up Node
+      uses: actions/setup-node@v6
+      with:
+        node-version-file: package.json
+        cache: yarn
+        registry-url: https://npm.manageiq.org/
+    - name: Install dependencies
+      run: yarn install
+    - name: Extract translations
+      run: yarn gettext:extract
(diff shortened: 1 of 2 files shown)


.github/workflows/locale_update.yaml
```diff
@@ -0,0 +1,51 @@
...
+    if: github.repository_owner == 'ManageIQ'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Set up Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: '22'
+    - name: Install dependencies
+      run: yarn install
```
Fryguy, line 26: This should be combined together (better caching) and use the node version file so we don't have to maintain the version number in multiple places. See the ci.yaml


```suggestion
    - name: Set up Node
      uses: actions/setup-node@v6
      with:
        node-version-file: package.json
        cache: yarn
        registry-url: https://npm.manageiq.org/
```

Fryguy: Also setup-node v4 is super-mega-old

.github/workflows/locale_update.yaml
```diff
@@ -0,0 +1,51 @@
...
+  cancel-in-progress: true
+permissions:
+  contents: read
+jobs:
+  locale_update:
+    if: github.repository_owner == 'ManageIQ'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Set up Node
```
Fryguy, line 19: We might need the systme setup before the node setup, like in the ci.yaml


```suggestion
    - name: Set up system
      run: bin/before_install
    - name: Set up Node
```

jrafanie: done

.github/workflows/locale_update.yaml
jrafanie: done

jrafanie:
I made the changes and quoted the cron string.  We can see how it works once it's merged.

